PR #221, opened 2017-12-28
Update markdown to support CommonMark

qinezh:
Before:
image

After:
image


alan-geller:
I believe this PR is now obsolete. The live site looks like the "after" image above.

PR #248, opened 2018-02-26
Fix ordering of bullets in install guide.

cgranade:


martinquantum:
I just found that one too, thanks for being a few steps ahead of me, Chris :)

PR #253, opened 2018-04-26
Conceptual docs edits

mcurmei627:


cgranade:
Thank you so much for the help, these look like great edits!

PR #261, opened 2018-05-25
Typo fixes in AssertProb and Measure

frtibble:
This fixes the following feedback from Balint Rikker:

1. Typo in the documentation for AssertProb:
https://quantum.uservoice.com/forums/906946-samples-and-documentation/suggestions/33472849-incorrect-example-in-assertprob-documentation

2. Typos in documentation for Measure:
https://docs.microsoft.com/en-us/qsharp/api/prelude/microsoft.quantum.primitive.measure?view=qsharp-preview
In the formula for the probability distribution, the tensor product of the Pauli bases should be indexed from either 0 to N-1, or from 1 to N as there are only N of them.

P0⊗P1⊗⋯⊗PN

On the line below, the explanation defines 'N=Length(paulis)' but instead of 'paulis', it should be 'bases', to match the type signature below.

### DIFF
--- a/api/prelude/microsoft.quantum.primitive.measure.yml
+++ b/api/prelude/microsoft.quantum.primitive.measure.yml
@@ -12,13 +12,13 @@ summary: >2
            \frac12 \braket{
                \psi \mid|
                \left(
-                   \boldone + P_0 \otimes P_1 \otimes \cdots \otimes P_N
+                   \boldone + P_0 \otimes P_1 \otimes \cdots \otimes P_N-1
                \right) \mid|
                \psi
            },
    \end{align}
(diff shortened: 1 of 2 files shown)


api/prelude/microsoft.quantum.primitive.measure.yml
```diff
@@ -12,13 +12,13 @@ summary: >2
            \frac12 \braket{
                \psi \mid|
                \left(
-                   \boldone + P_0 \otimes P_1 \otimes \cdots \otimes P_N
+                   \boldone + P_0 \otimes P_1 \otimes \cdots \otimes P_N-1
```
tcNickolas: Should this be P_{N-1}?

PR #264, opened 2018-06-04
Typo fix in Amplitude Amplification Oracles

frtibble:
Found by Balint Rikker:
"On page https://docs.microsoft.com/en-us/quantum/libraries/data-structures?view=qsharp-preview : 

'This oracle is then a special case of the RAll1 operation, which allows for rotating by an arbitrary phase instead of the reflection case  ϕ=−1.'

The value of phi for the reflection case should be pi instead of -1.
The confusion probably comes from the fact that calling RAll1(PI(), qubits) results in the amplitude of the |11..1> state being multiplied by -1 (i.e. being reflected)."


cgranade:
I believe that you are correct, thank you so much for catching that mistake!

PR #272, opened 2018-06-30
Anpaz/update qdk version

anpaz-msft:


anpaz-msft:
I guess we need to merge from master.

PR #274, opened 2018-07-08
Improved install guide

cgranade:


articles/quantum-InstallConfig.md
```diff
@@ -12,19 +12,16 @@ ms.topic: article
...
-## Required Prerequisites
+The Quantum Development Kit can be used either with the Visual Studio 2017 integrated development environment, or with the command line and a development editor such as Visual Studio Code.
+We'll walk through both methods in this install guide.
 
-- A 64-bit installation of Windows, macOS, or Linux. (the Microsoft Quantum Development Kit has been tested under Ubuntu Linux, but may work on other distributions).
-- The [.NET Core SDK 2.0](https://www.microsoft.com/net/learn/get-started) or later.
+> [!NOTE]
+> The Quantum Development Kit requires a 64-bit installation of Windows 10, macOS, or Linux.
 
+### Optional Prerequisites
```
tcNickolas: I would suggest moving this section to after installing IDE and QDK; since they are optional, they should be lower priority than the main workflow

articles/quantum-InstallConfig.md
```diff
@@ -33,34 +30,54 @@ ms.topic: article
...
+### [Installing Visual Studio Code and the .NET Core SDK (Windows, macOS, Linux)](#tab/tabid-vscode)
 
-#### [Installing Visual Studio Code (Windows, macOS, Linux)](#tab/tabid-vscode)
+The Quantum Development Kit uses the .NET Core SDK (2.0 or later) to make it easy to create, build, and run Q# projects from the command line.
+We recommend using the Quantum Development Kit together with Visual Studio Code.
 
-You can download Visual Studio Code for free for your platform.
+1. Please see [Getting started with .NET](https://www.microsoft.com/net/learn/get-started) for how to download and install the .NET Core SDK.
+2. Once the .NET Core SDK is installed, run the following command in your favorite command line (e.g.: PowerShell or Bash) to download the latest templates for creating new Q# applications and libraries:
+   ```bash
```
tcNickolas: This code snippet doesn't render right in preview

articles/quantum-InstallConfig.md
```diff
@@ -33,34 +30,54 @@ ms.topic: article
...
+   $ dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.2.1806.3001-preview"
+   ```
 1. Go to the [Visual Studio Code website](https://code.visualstudio.com/).
 2. Pick the download for your platform and follow the installation prompts.
+    > [!NOTE]
```
tcNickolas: This note doesn't render right in preview

tcNickolas:
"Install libraries for non-commercial use" instruction looks a bit weird - it's only mentioned in VS 2017 guide, and if somebody follows the instruction precisely, they can't run teleport because they have Quantum-NC.sln open.

tcNickolas:
VS 2017 guide mentions that you'll be prompted to install F# among missing features, but VS Code only mentions QDK and C# - does VS Code not want to install F#?

articles/quantum-InstallConfig.md
```diff
@@ -33,34 +30,54 @@ ms.topic: article
...
+
+> [!TIP]
+> You can also install the Quantum Development Kit extension for Visual Studio Code from the *Go to file...* palette.
+> Press **Ctrl+P** or **⌘+P** from within Visual Studio Code, type or paste `ext install quantum.quantum-devkit-vscode`, and press **Enter**.
 
 ****
 
-## Installing and Validating the Q# Development Environment
+You should now have the Quantum Development Kit installed and ready to use in your own applications and libraries.
+The next section will cover how to build the samples and standard libraries provided along with the Quantum Development Kit.
```
tcNickolas: Is this phrase redundant? It just repeats the next section tile

cgranade: I think it helps the prose read better across that transition, but I can reword it, then.

cgranade:
@tcNickolas: Thanks for looking over it, I think I've addressed your comments now. Please let me know if you're good for me to merge this in. Thanks!

With respect to F# and VS Code, the .NET Core SDK already provides F# support. If a user wants to use F# from within VS Code, then the Ionide extensions work quite well for that. By contrast, F# compilation and editor support are coupled in VS2017, so you kind of need for that to be there.

articles/quantum-InstallConfig.md
```diff
@@ -126,3 +132,15 @@ In this section you will clone the quantum samples & libraries repository, and r
...
+### Optional Dependencies ###
+
+Some of the samples demonstrate using the Quantum Development Kit with other programming environments, and have additional installation requirement:
+
+- Some samples use the Node.js Package Manager (NPM) to handle user interface dependencies.
+  NPM can be installed [manually](https://nodejs.org/en/download/current/) or through a [package manager](https://nodejs.org/en/download/package-manager/).
+- The Python interoperability feature has been developed for the [Anaconda distribution](https://www.anaconda.com/download/) of Python 3.6.
+  Please see the [README](https://github.com/Microsoft/Quantum/tree/master/Samples/PythonInterop/README.md) file provided with the Python sample for more details.
+
+Additionally, the Quantum Development Kit is provided along with libraries for non-commercial use.
```
tcNickolas: Is QDK actually provided with these libraries, or are these libraries provided separately? I suspect there is a line to be drawn here. I'd say something like "You can find more Q# libraries for non-commercial use at ..."

articles/quantum-InstallConfig.md
tcNickolas: I still don't like it... Together with the header and the first sentence of the next paragraph it's repeated three times.

articles/quantum-InstallConfig.md
cgranade: Since the non-commercial phase estimation operations are documented with the rest of the QDK APIs, I think it makes sense to keep the phrasing as is for now, with an eye to revisiting as you suggest.

PR #277, opened 2018-07-12
Add article on Quantum oracles to Concepts

tcNickolas:


articles/TOC.yml
```diff
@@ -17,6 +17,8 @@
       href: quantum-concepts-7-PauliMeasurements.md
     - name: Quantum circuits
       href: quantum-concepts-8-QuantumCircuits.md
+    - name: Quantum oracles
```
cgranade: Looks like the numbers are out of order here, might be good in the future to remove the numbers from the URLs but I don't think it's a problem for now.

PR #282, opened 2018-08-01
Split install guide into Windows 10 and cross-platform pages.

cgranade:
This PR addresses a common confusion about which set of install instructions to follow by clarifying that the Visual Studio 2017 instructions are Windows 10 only, and that the Visual Studio Code / command-line instructions are cross-platform. In particular, the install guide is now separated entirely into two distinct pages, with a selection page that presents readers with screenshots from supported operating systems to emphasize the split.

Still TODO: emphasize range of supported versions for VS 2017, and explain the difference between VS 2017 and VS Code.


articles/install-guide/vs-2017.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Installing the Quantum Development Kit for Visual Studio 2017 #
+
+## Installing Visual Studio 2017 (Windows 10 only)
+
+> [!NOTE]
+> The Microsoft Quantum Development Kit does not support Visual Studio for Mac.
+> To install on macOS, please follow the instructions for [using Visual Studio Code](xref:microsoft.quantum.install.vs-code).
+
+If you do not have Visual Studio installed, you can download Visual Studio 2017 Community Edition for free.
```
anpaz-msft: Instead of providing instructions why don't we just point people to this URL:
https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio

They are better documented and easier to follow, and owned by VS.

If we do, there are two important things need to highlight:
1. People need to have VS 2014 Update 4 or newer (older version will not work with .net core)
2. People need to select the **.NET Core cross-platform development** workload.

anpaz-msft: If we point to the URL, installation for Windows 10 consists only on installing the VSIX.

articles/install-guide/vs-2017.md
```diff
@@ -0,0 +1,71 @@
...
+1. Go to the [Visual Studio download page](https://www.visualstudio.com/downloads/).
+1. Click on the Visual Studio Community **Free download** button.
+2. Navigate to your browser's download folder and double click on the executable file whose name begins with **vs_community**. The file name will contain a sequence of numbers that varies.
+3. _**Important!**_ &nbsp;When you are presented with the option to select the tools for specific workloads, check the box for **.NET Core cross-platform development**.
+   If there are any other workloads you would like to install, you can select them as well at this step.
+4. After selecting your workloads, click **Install** to complete the installation.
+5. Once you have installed Visual Studio 2017, download the [Microsoft Quantum Development Kit for Visual Studio](https://marketplace.visualstudio.com/items?itemName=quantum.DevKit) extension from the Visual Studio Marketplace.
+   Double-click the VSIX once it is downloaded to install the extension.
+
+## Building the Quantum Samples and Libraries
```
anpaz-msft: This is not part of the installation, so I wouldn't include it in the "Installing the QDK" section.

I would make this a new seccion:
\# Validating the Quantum Development Kit installation

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Installing the Q# development environment for VS Code
+author: chgranad
+uid: microsoft.quantum.install.vs-code
+ms.author: christopher.granade@microsoft.com
+ms.date: 7/31/2018
+ms.topic: article
+---
+
+# Installing the Quantum Development Kit for Visual Studio Code #
```
anpaz-msft: Instead of installing on Visual Studio Code. I would give instructions on how to do this from the command line. People get confused and think they need VS Code to work.

Then, at the end I would recommend installing VS Code and the Extension.

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
...
+
+The Quantum Development Kit uses the .NET Core SDK (2.0 or later) to make it easy to create, build, and run Q# projects from the command line.
+We recommend using the Quantum Development Kit together with Visual Studio Code.
+
+1. Please see [Getting started with .NET](https://www.microsoft.com/net/learn/get-started) for how to download and install the .NET Core SDK.
```
anpaz-msft: As you point out, some people gets confused when we send them to "Getting started with .NET". Why don't we point them instead to:
https://www.microsoft.com/net/download
and ask them to download the **.NET Core SDK**

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
...
+
+2. Once the .NET Core SDK is installed, run the following command in your favorite command line (e.g.: PowerShell or Bash) to download the latest templates for creating new Q# applications and libraries:
+   ```Bash
+   $ dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.2.1806.3001-preview"
+   ```
+
+3. Go to the [Visual Studio Code website](https://code.visualstudio.com/).
```
anpaz-msft: Again, I would move installing VS Code and the extension as an optional sub-section.

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
...
+> When you run Visual Studio Code for the first time, you may be prompted to install Git if you have not already done so.
+> Git is not required to use the Quantum Development Kit, but works great for managing and sharing Q# projects, and for contributing back to the Q# libraries and samples.
+
+> [!TIP]
+> You can also install the Quantum Development Kit extension for Visual Studio Code from the *Go to file...* palette.
+> Press **Ctrl+P** or **⌘+P** from within Visual Studio Code, type or paste `ext install quantum.quantum-devkit-vscode`, and press **Enter**.
+
+You should now have the Quantum Development Kit installed and ready to use in your own applications and libraries.
+
+## Building the Quantum Samples and Libraries
```
anpaz-msft: As with VS installation. Move this to its own "Validating" section, instead of a sub-section of "installing"

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
...
+
+In this section you will clone the quantum samples & libraries repository, and run a sample application to verify that your Q# environment is correctly installed and configured.
```
anpaz-msft: We need to add a pointer on how to install git.

articles/install-guide/vs-code.md
```diff
@@ -0,0 +1,90 @@
...
+
+3. Clone and open the [Microsoft Quantum Developer Kit Samples and Libraries](https://github.com/microsoft/quantum) GitHub repository.
+  ```bash
+  $ git clone https://github.com/Microsoft/Quantum.git
+  ```
+  To continue at the command line, navigate into the newly cloned directory:
+  ```bash
+  $ cd Quantum
+  ```
+  If you would like to work with the new project in VS Code, run:
```
anpaz-msft: I would skip this as it's not necessary.

articles/quantum-InstallConfig.md
```diff
@@ -15,131 +15,28 @@ ms.topic: article
...
+[<img src="media/install-guide-select-vs-2017.png" width="90%">](xref:microsoft.quantum.install.vs-2017)
+<div class="nextstepaction">
+[Install for Visual Studio 2017 on Windows 10](xref:microsoft.quantum.install.vs-2017)
+</div>
+</td>
+
+<td>
+[<img src="media/install-guide-select-vs-code.png" width="90%">](xref:microsoft.quantum.install.vs-code)
+<div class="nextstepaction">
+[Install for Visual Studio Code on Windows 10, macOS, or Linux](xref:microsoft.quantum.install.vs-code)
```
anpaz-msft: I would rename this to "Install from the command line on Windows 10, macOS or Linux"

articles/quantum-InstallConfig.md
```diff
@@ -15,131 +15,28 @@ ms.topic: article
...
 
 The Quantum Development Kit can be used either with the Visual Studio 2017 integrated development environment, or with the command line and a development editor such as Visual Studio Code.
-We'll walk through both methods in this install guide.
+
+<table>
+<tr>
+<td>
+[<img src="media/install-guide-select-vs-2017.png" width="90%">](xref:microsoft.quantum.install.vs-2017)
+<div class="nextstepaction">
+[Install for Visual Studio 2017 on Windows 10](xref:microsoft.quantum.install.vs-2017)
```
anpaz-msft: Install *for*, or Install *on*?

articles/quantum-InstallConfig.md
```diff
@@ -15,131 +15,28 @@ ms.topic: article
...
+</div>
+</td>
+
+<td>
+[<img src="media/install-guide-select-vs-code.png" width="90%">](xref:microsoft.quantum.install.vs-code)
```
anpaz-msft: On the same spirit, I would replace the screenshot with one that shows a command line on the front and VS-code on the back.

anpaz-msft:
+Krysta as she just went through this and might have some more feedback.

articles/install-guide/vs-2017.md
cgranade: Fair point, I'm definitely happy to split that out. Many users have been confused by the inclusion of the validation step as a part of the installation guide, after all.

articles/install-guide/vs-code.md
cgranade: Sounds good, I'll change that accordingly, then.

articles/install-guide/vs-code.md
cgranade: That part has been helpful in the past, especially for users not used to the anycode/Open Folder workflow. While VS Code acts similarly to Sublime, Atom, and other text editors in that regards, many users aren't familiar with text editors in general. From that perspective, this line can help people get started.

articles/quantum-InstallConfig.md
cgranade: I didn't want to use the word "on" in both places, but I agree, not quite right yet. I'll reword accordingly, then.

articles/install-guide/vs-code.md
cpalmer2020: Can you remove the "$" from the bash code box so that a copy/paste will work?

bettinaheim: Same comment as above.

articles/install-guide/vs-code.md
bettinaheim: I don't think there is any harm in doing this additional step. Maybe a sentence on what it does and that it's optional may be useful?

articles/quantum-InstallConfig.md
bettinaheim: Why exactly are we so eager about not having VS Code mentioned in a prominent place? I think it should at least have equal visibility as VS.

articles/quantum-InstallConfig.md
cgranade: If it's alright, I'd like to defer the screenshot, since that could make the call to action a bit busy graphically speaking.

cgranade:
Thanks for the feedback, @anpaz-msft, and my apologies for dropping the ball. At any rate, I think that other than the screenshot suggestion, adding a third split to the install guide should help make it more explicit that we support the command line in whatever editor a user might prefer, while still focusing on our efforts to provide good integration for VS Code. If you could give it another look, I'd really appreciate it. Thanks!

cgranade:
My apologies for forgetting to add the new page to the TOC, that should be fixed now.

PR #301, opened 2018-10-30
Adding links to github

anpaz-msft:
Adding a couple of links from the homepage to our open-source repos to send more traffic to them, specially QuantumLibraries since we don't need talk much about it anywhere else as people only need to clone /Quantum to get their samples and validate their installation.


tcNickolas:
Should we include a link to the katas repo as well? Something along the lines of "If you want to dive deeper into Q# programming, check out the Quantum Katas - a collection of self-paced tutorials introducing you to quantum computing via programming exercises in Q#."?

PR #306, opened 2018-10-31
chore: Replace ~ images with relative links for GH

nschonni:
These links are OK for DocFx, but GitHub doesn't like them. Should get ride of some build warnings too.


nschonni:
Hmm, still unhappy on the build, so i'll close for now

PR #322, opened 2018-11-06
Missing $ for inline equation on ContinuousOracle type

mlxd:
The documentation for the ContinuousOracle type (https://docs.microsoft.com/en-us/qsharp/api/canon/microsoft.quantum.canon.continuousoracle?view=qsharp-preview) has a missing $ symbol to close the inline equation. This change is modified in QuantumLibraries as well as here.


anpaz-msft:
Thanks! The fix was picked up after we ran the tool to generate the API docs for the latest release, so closing this PR.

PR #339, opened 2019-01-03
Change $N_{ij} = M^*_{ji} to $N_{ij} = M_{ji}^*

lwoodyiii:
[Vectors and Matrices page](https://docs.microsoft.com/en-us/quantum/concepts/vectors-and-matrices?view=qsharp-preview) is not formatted correctly and fails at the paragraph "For any matrix M,....". Hopefully this will fix it. I am not an expert in Math markdown :(


lwoodyiii:
Cool!

PR #388, opened 2019-04-23
Add headings to C# quickstart for SEO.

MikeDodaro:


articles/quickstart.md
```diff
@@ -56,10 +56,12 @@ $ code . # To open in Visual Studio Code.
 
 Our goal is to create a [Bell State](https://en.wikipedia.org/wiki/Bell_state) showing entanglement. We will build this up piece by piece to show the concepts of qubit state, gates and measurement.
 
+#### Source files of a Quantum application
```
cgranade:
```suggestion
#### Source files for a quantum application
```

articles/quickstart.md
```diff
@@ -256,6 +262,7 @@ namespace Quantum.Bell
 }
 ```
 
+#### Run Quantum operations from C# driver
```
cgranade: We've up until this point been fairly inconsistent with respect to driver vs host program, but I think it'd be good if we could use the latter in preference to the former going forward.
```suggestion
#### Run quantum operations from a C# host program
```

PR #390, opened 2019-04-24
Update set-array syntax.

MikeDodaro:
Update set-array syntax in topic: Invalidated Qubits Use Checker

### DIFF
--- a/articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
+++ b/articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
@@ -17,7 +17,7 @@ If there is only one body and no other specializations, one can specify the impl
 ```qsharp
 operation UseReleasedQubitTest () : Unit {
     mutable q = new Qubit[1];
-    using(ans = Qubit[1]) {
+    using(ans = Qubit()) {
         set q w/= 0 <- ans[0];
     }
     H(q[0]);
--- a/articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
+++ b/articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
@@ -18,7 +18,7 @@ If there is only one body and no other specializations, one can specify the impl
 operation UseReleasedQubitTest () : Unit {
     mutable q = new Qubit[1];
     using(ans = Qubit()) {
-        set q w/= 0 <- ans[0];
+        set q w/= 0 <- ans;
     }
     H(q[0]);
 }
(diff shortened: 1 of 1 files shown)


articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
```diff
@@ -20,7 +20,7 @@ operation UseReleasedQubitTest () : Unit {
     body (...) {
```
bettinaheim: The change is correct. going further than that, if there is only one body (and no other specializations), one can specify the implementation directly within the operation declaration. Additionally, single qubits can be allocated by now (both of these things are optional, but it would be good for docs to illustrate how code should be written). with that the example then becomes

operation UseReleasedQubitTest () : Unit {
    mutable q = new Qubit[1];
    using(ans = Qubit()) {
        set q w/= 0 <- ans;
    }
    H(q[0]);
}

articles/machines/qc-trace-simulator/invalidated-qubits-use-checker.md
```diff
@@ -10,20 +10,17 @@ uid: microsoft.quantum.machines.qc-trace-simulator.invalidated-qubits
 
 # Invalidated Qubits Use Checker
 
-The `Invalidated Qubits Use Checker` is a part of the quantum computer [Trace
-Simulator](xref:microsoft.quantum.machines.qc-trace-simulator.intro) designed for detecting
-potential bugs in the code. Consider the following piece of Q# code to
-illustrate the issues detected by the `Invalidated Qubits Use Checker`.
+The `Invalidated Qubits Use Checker` is a part of the quantum computer [TraceSimulator](xref:microsoft.quantum.machines.qc-trace-simulator.intro) designed for detecting potential bugs in the code. Consider the following piece of Q# code to illustrate the issues detected by the `Invalidated Qubits Use Checker`.
+
+If there is only one body and no other specializations, one can specify the implementation directly within the operation declaration. Additionally, single qubits can be allocated (both of these are optional but can be useful).
```
bettinaheim: In case that wasn't clear: I didn't mean for this explanation to go in here - the explanation was meant for you regarding what could be updated in code samples in general.

MikeDodaro: Thanks, Bettina.  I suppose it should be included in this topic: https://docs.microsoft.com/en-us/quantum/language/type-model?view=qsharp-preview#type-parameterized-functions-and-operations

bettinaheim: I think we more or less have these things in the docs (e.g. here for the using statement: https://docs.microsoft.com/en-us/quantum/language/statements?view=qsharp-preview#qubit-management and here for specializations: https://docs.microsoft.com/en-us/quantum/language/file-structure?view=qsharp-preview#operation-definitions). 
So in that sense, I was mostly just trying to be helpful, because while we do our best to adapt documentation when new features come in, sometimes code examples in particular do not get updated to make use of them.

bettinaheim: So, in that sense - when you come across such a code example, it would be awesome if you could update it. :)

PR #393, opened 2019-04-24
Attempt to correct document math display error.

MikeDodaro:


tcNickolas:
This is an improvement but doesn't fix the whole section. 

image

I suspect that "=" on a separate line makes the lines above be treated as a header, but I haven't verified it.

PR #401, opened 2019-04-25
Cannon -> Standard

MikeDodaro:


bettinaheim:
Leaving that to @cgranade to say it this makes sense to change - Canon still exists, and our template currently import Intrinsics and Canon by default, but not Standard. Chris, if Standard is the better default choice, should be adapt the templates, or is Intrinsics + Canon a good default?

articles/install-guide/python.md
```diff
@@ -49,7 +49,7 @@ To get started, create a new `Quantum.qs` file with the following content:
 namespace Microsoft.Samples 
 {
     open Microsoft.Quantum.Primitive;
-    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Standard;
```
cgranade:
```suggestion
```

PR #411, opened 2019-05-01
Move non software stack text from Software stack article  to new topic.

MikeDodaro:


MikeDodaro:
I can't see the review builds on my account, but I think the warnings showing here do not reflect the display of the image: [Warning] Invalid file link:(~/media/concepts_stack.png).  The diagram does show.

PR #417, opened 2019-05-13
Adding docs for the Numerics library.

thomashaener:


articles/libraries/numerics/numerics.md
```diff
@@ -0,0 +1,148 @@
...
+        
+        // ... (use the result)
+    }
+}
+```
+
+## Sample: Evaluating smooth functions
+
+To evaluate smooth functions such as $\sin(x)$ on a quantum computer, where $x$ is a quantum `FixedPoint` number,
+the QDK Numerics Library provides the operations `EvaluatePolynomialFxP` and `Evaluate[Even/Odd]PolynomialFxP`.
```
cgranade: These as well would be good candidates for linking to API references.
```suggestion
the Quantum Development Kit numerics library provides the operations `EvaluatePolynomialFxP` and `Evaluate[Even/Odd]PolynomialFxP`.
```

articles/libraries/numerics/numerics.md
```diff
@@ -0,0 +1,148 @@
...
+
+```bash
+git clone https://github.com/Microsoft/Quantum.git
+cd Quantum/Numerics
+```
+
+Then, `cd` into one of the sample folders and run the sample via
+
+```bash
+cd [SampleFolder]
```
cgranade: Readers still learning how to navigate the command line may think that this is literal syntax; could we perhaps do something like give a concrete example, and say "for example" at L143?

PR #434, opened 2019-05-31
Updating with v0.7 docs.

anpaz-msft:


cgranade:
> The Q# chemistry docs look good. However, the C# chemistry component references seems to still refer to the v0.6 version.

That makes sense, I think, as this PR only updates the Q# API references.

PR #456, opened 2019-07-11
Add Microsoft Quantum brand

cjgronlund:
Add brand to articles
Add language and brand to Q# ref docs

Global settings in docfx file

### DIFF
--- a/api/docfx.json
+++ b/api/docfx.json
@@ -56,8 +56,7 @@
     },
     "titleSuffix": {
       "*.md": "Q# reference - Microsoft Quantum"
-      }
-    },
+      },
     "fileMetadata": {},
     "template": [],
     "dest": "quantum-api"
--- a/api/docfx.json
+++ b/api/docfx.json
@@ -54,6 +54,10 @@
       "feedback_github_repo": "Microsoft/quantum",
       "feedback_product_url": "https://github.com/microsoft/quantum/issues"
     },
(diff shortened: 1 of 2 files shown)


api/docfx.json
```diff
@@ -54,6 +54,10 @@
       "feedback_github_repo": "Microsoft/quantum",
       "feedback_product_url": "https://github.com/microsoft/quantum/issues"
     },
+    "titleSuffix": {
+      "*.md": "Q# reference - Microsoft Quantum"
+      }
```
cgranade: It looks like there is an extra `}` here that is preventing DocFX from provisioning the docset correctly.
```suggestion
```

articles/docfx.json
```diff
@@ -64,6 +64,10 @@
         }
       }
     },
+    "titleSuffix": {
+      "*.md": "Microsoft Quantum"
+      }
```
cgranade: Similarly here:
```suggestion
```

api/docfx.json
cjgronlund: Thanks for the heads up. I'll fix and resubmit.

articles/docfx.json
cjgronlund: Validation complete and fix submitted. Thank you!

cgranade:
Wonderful, thanks for the help, @cjgronlund!

PR #472, opened 2019-07-22
Add quantum brand - attempt #2

cjgronlund:


cgranade:
The title suffix doesn't seem to show up on the review site its under `globalMetadata`, but I tried putting it under `fileMetadata` instead and that seems to work:
```
"fileMetadata": {
      "titleSuffix": {
        "**/**.md": "Microsoft Quantum"
      },
      // MathJaX config here.
}
```

cjgronlund:
@cgranade - Thanks for the heads up. The fileMetadata setting: Have you already tested this setting? I'm unable to see a preview build of the PR I submitted. But, I'm following this example: https://github.com/MicrosoftDocs/sql-docs-pr/blob/master/docs/docfx.json

cgranade:
@cjgronlund: No worries, thanks for helping us get the title suffix working, I didn't even know there was such a setting until your first PR! It looks from the review rendering like the `fileMetadata` approach works, but not sure why it would be any different than `globalMetadata`...

cjgronlund:
@cgranade Thanks for the heads up. I was having trouble seeing the staged version, so thanks for checking. I'll make the change and resubmit. Interesting how this is done different ways in different repos!

cjgronlund:
@cgranade: I think this should do it. Thank you for your feedback. :>)

PR #476, opened 2019-08-05
Update metadata for Quantum docs reporting

natke:


natke:
Thank you for the approval @cgranade. Can you please merge the PR?

(We should also fix the build warnings, in a separate PR)

PR #486, opened 2019-08-14
Q# manual (merger of Q# P. Language and Q. Dev. Techniques) (Branch, not fork)

geduardo:
I create a new PR but from a branch of the main repo. I deleted the old one. The content is the same but with some fixes thanks to the preview mode. I make this PR because I want the bots to point out broken xrefs, links, etc. Also to help @bromeg to suggest changes more easily.


crazy4pi314:
Just a comment for the team here, I am not sure if the "View" links in the Preview URL column of the build reports above are supposed to be public. They currently are not, and I get a 403 error (see below). I can always look at the raw but it would be nice to see if formatted if that was possible :) 

image

geduardo:
@crazy4pi314 currently you need an authorized "...@microsoft.com" user to see the preview url :(
I don't know if that can be changed easily

articles/manual/basics.md
```diff
@@ -1,22 +1,15 @@
 ---
-# Mandatory fields. See more on aka.ms/skyeye/meta.
-title: Quantum development techniques introduction | Microsoft Docs 
-description: Quantum development techniques introduction
-author: QuantumWriter
-ms.author: Christopher.Granade@microsoft.com
-ms.date: 12/11/2017
+title: The basics of Q# | Microsoft Docs
```
cgranade: It'd be good to check on the preview site, but I don't think we need the `| Microsoft Docs` part now that that's in the global metadata.

articles/manual/basics.md
```diff
@@ -44,15 +37,27 @@ In this way, Q# makes it easy to express the logic underlying quantum and hybrid
 
 ## Q# operations and functions
 
-Concretely, a Q# program is comprised of one or more *operations*, one or more *functions*, and user defined types. Operations are used to describe the transformations of the state of a quantum machine and are the most basic building block of Q# programs. 
-Each operation defined in Q# may then call any number of other operations, including the built-in *intrinsic* operations implemented by a target machine.
-When compiled, each operation is represented as a .NET class type that can be provided to target machines.
+ Concretely, a Q# program is comprised of one or more *operations*, one or more *functions*, and user defined types. Operations are used to describe the transformations of the state of a quantum machine and are the most basic building block of Q# programs. 
+ Each operation defined in Q# may then call any number of other operations, including the built-in *intrinsic* operations implemented by a target machine.
+ When compiled, each operation is represented as a .NET class type that can be provided to target machines.
```
cgranade: I recognize that this was there in the original, but it may be good to remove this sentence now that we support Python as a host language as well as .NET languages.

articles/manual/basics.md
```diff
@@ -44,15 +37,27 @@ In this way, Q# makes it easy to express the logic underlying quantum and hybrid
...
+ 
+### Expressions
+An expression in a programming language is a combination of one or more constants, variables, operators, and functions that the programming language interprets and computes to produce another value. For example, `2+3` is an arithmetic and programming expression which evaluates to `5`. Q# has its own expressions and rules.
+### Statements 
+  A statement is a syntactic unit of an imperative programming language that expresses some action to be carried out. Statements contrast with expressions in that statements do not return results and are executed solely for their side effects, while expressions always return a result and often do not have side effects at all. This distinction is frequently observed in wording: a statement is executed, while an expression is evaluated. An example of a Q# statement is the loop `for` which supports iteration over an integer range.
 
-## Structure of Q# Source Files
+## Structure fo Q# Source Files
```
cgranade:
```suggestion
## Structure of Q# Source Files
```

articles/manual/index.md
```diff
@@ -0,0 +1,95 @@
...
+A natural model for quantum computation is to treat the quantum computer
+as a coprocessor, similar to that used for GPUs, FPGAs, and other adjunct
+processors.
+The primary control logic runs classical code on a classical "host" computer.
+When appropriate and necessary, the host program can invoke a subroutine
+that runs on the adjunct processor.
+When the subroutine completes, the host program gets access to the
+subroutine's results.
+
+Q# provides a small set of primitive types, along with two ways
```
cgranade:
```suggestion
Q# provides a small set of built-in types, along with two ways
```
There's also UDTs in addition to arrays and tuples.

articles/manual/index.md
```diff
@@ -0,0 +1,95 @@
...
+(arrays and tuples) for creating new, structured types.
+It supports a basic procedural model for writing programs,
```
cgranade: I think this sentence made a bit more sense in the original version than it does here; we might not want to open by implying that Q# only supports a basic model, as that works against teaching the reader about the high-level functionality offered by Q#.

articles/manual/index.md
```diff
@@ -0,0 +1,95 @@
...
+with loops and if/then statements.
+The top-level constructs in Q# are user defined types, operations,
+and functions.
+
+Q# is scalable: it can be used to write simple demonstration programs like teleport that run on a few qubits, but also supports writing large, sophisticated programs such as simulations of complex molecules that will require large machines with millions of qubits. Even though large physical machines are still in the future, Q# allows a programmer to program complex quantum algorithms now. What is more, Q# supports various tasks such as debugging, profiling, resource estimation, and certain special-purpose simulations in a scalable way.
+
+## Why Q#?
+
+[![Quantum computing video](http://img.youtube.com/vi/cOUrzxyng04/0.jpg)](http://www.youtube.com/watch?v=cOUrzxyng04 "")
```
cgranade: This video isn't showing up as a video for some reason, such that it isn't obvious that I should be able to click through and watch it. It may also be good to check if there's a more up-to-date video that we can use; @bromeg may have some resources to suggest there.

articles/manual/index.md
```diff
@@ -0,0 +1,95 @@
...
+ While quantum hardware is still in early development, we need to build software for quantum computers to take advantage
+ of the quantum power of computation when the time comes. Q# is a tool made to provide a platform where developers can write and test programs to be run on a future generation of quantum computers.
+
+For more information about the foundations and motivation of Q# check out [Why do we need Q#?](https://devblogs.microsoft.com/qsharp/why-do-we-need-q/).
+
+## How to learn Q#?
+
+### Pre-required knowledge to use Q#
+
+There is no need to be an expert on quantum physics to start writing your first quantum programs. However, there are some concepts that are essential to understand quantum computing:
```
cgranade: I think this could read in a way that scares off new readers; there's plenty of room to learn by doing, especially with the katas, books, courses, and so forth that are now available. Especially since you do a great job of listing those below, it might be good to invite the reader to keep reading even if they don't yet have all the background they will eventually need.

articles/manual/index.md
```diff
@@ -0,0 +1,95 @@
...
+  have a deep understanding of quantum mechanics to use Q#, it is highly
+  recommended to be familiar with some basic concepts, namely
+  *quantum state*, *qubit*, *observable*, *measurement*, *Hamiltonian*,
+  *superposition*, *entanglement*, *interference*, among others.  
+
+### Learning materials
+
+Quantum computing is an emerging field and fortunately, there is an immense amount of learning resources online. Sometimes so many options can get overwhelming, so here we gather a few of them.
+
+#### Quantum computing concepts
```
cgranade: I think this could be combined with the section starting at line 84, as learning Q# often includes learning about quantum computing more generally as well.

articles/manual/basics.md
```diff
@@ -44,15 +37,27 @@ In this way, Q# makes it easy to express the logic underlying quantum and hybrid
...
-When compiled, each operation is represented as a .NET class type that can be provided to target machines.
+ Concretely, a Q# program is comprised of one or more *operations*, one or more *functions*, and user defined types. Operations are used to describe the transformations of the state of a quantum machine and are the most basic building block of Q# programs. 
+ Each operation defined in Q# may then call any number of other operations, including the built-in *intrinsic* operations implemented by a target machine.
+ When compiled, each operation is represented as a .NET class type that can be provided to target machines.
+
+ In contrast to operations, functions are used to describe purely classical behavior and do not have any effects besides computing classical output values. Q# is a strongly typed language and comes with a set of built-in primitive types as well as support for user defined types. 
+
+ Throughout the rest of this guide, we will see how to use different language concepts and constructs to help us define complex quantum programs through the basic building blocks of operations, functions, and types.
+
+## Q# syntax overview
```
cgranade: This feels a bit out of place now that this content is coming a lot earlier in the documentation than it used to. It might work better to show by example, returning to the `H` example at the top of the file, but as a complete Q# source file.

articles/manual/operations-functions.md
```diff
@@ -283,3 +248,119 @@ function SquareOperation(op : (Qubit => Unit)) : (Qubit => Unit) {
 
 In principle, the classical logic within `SquareOperation` could have been much more involved, but it is still isolated from the rest of an operation by the guarantees that the compiler can offer about functions.
 This approach will be used throughout the Q# standard library for expressing classical control flow in a way that can be readily used within quantum programs.
+
+## Generic Operations and Functions
```
cgranade:
```suggestion
## Type-Parameterized Operations and Functions
```

articles/manual/operations-functions.md
```diff
@@ -283,3 +248,119 @@ function SquareOperation(op : (Qubit => Unit)) : (Qubit => Unit) {
...
+> [!TIP]
+> This section assumes some basic familiarity with [generics in C#](https://docs.microsoft.com/dotnet/csharp/programming-guide/generics/introduction-to-generics), [generics in F#](https://docs.microsoft.com/dotnet/fsharp/language-reference/generics/), [C++ templates](https://docs.microsoft.com/cpp/cpp/templates-cpp), or similar approaches to metaprogramming in other languages.
+
+Many functions and operations that we might wish to define do not actually heavily rely on the types of their inputs, but rather only implicitly use their types via some other function or operation.
+For example, consider the *map* concept common to many functional languages; given a function $f(x)$ and a collection of values $\{x_1, x_2, \dots, x_n\}$, map returns a new collection $\{f(x_1), f(x_2), \dots, f(x_n)\}$.
+To implement this in Q#, we can take advantage of first class functions.
+Let's write out a quick example of `Map`, using ★ as a placeholder while we figure out what types we need.
+
+```qsharp
+function Map(fn : ★ -> ★, values : ★[]) : ★[] {
```
cgranade: It looks like this example is still using pre-0.6 names, it might be good to change this to be `Mapped` to clearly indicate that it is a function, and to parallel the name of the library function [Microsoft.Quantum.Arrays.Mapped](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.arrays.mapped).

articles/manual/operations-functions.md
```diff
@@ -283,3 +248,119 @@ function SquareOperation(op : (Qubit => Unit)) : (Qubit => Unit) {
...
+    return mappedValues;
+}
+```
+
+In principle, we could write a version of `Map` for every pair of types that we encounter, but this introduces a number of difficulties.
+For instance, if we find a bug in `Map`, then we must ensure that the fix is applied uniformly across all versions of `Map`.
+Moreover, if we construct a new tuple or UDT, then we must now also construct a new `Map` to go along with the new type.
+While this is tractable for a small number of such functions, as we collect more and more functions of the same form as `Map`, the cost of introducing new types becomes unreasonably large in fairly short order.
+
+Much of this difficulty results, from not having given the compiler the information it needs to recognize how the different versions of `Map` are related.
```
cgranade:
```suggestion
Much of this difficulty results from not having given the compiler the information it needs to recognize how the different versions of `Map` are related.
```

articles/manual/statements.md
```diff
@@ -1,14 +1,16 @@
...
 author: QuantumWriter
-uid: microsoft.quantum.qsharp-ref.statements
+uid: microsoft.quantum.manual.statements
 ms.author: Alan.Geller@microsoft.com 
 ms.date: 12/11/2017
 ms.topic: article
 ---
 
-# Statements and Other Constructs
+# Statements, Control Flow and other constructs
```
cgranade: The capitalization here should be consistent.
```suggestion
# Statements, Control Flow, and Other Constructs 
```

articles/manual/statements.md
```diff
@@ -1,14 +1,16 @@
...
+uid: microsoft.quantum.manual.statements
 ms.author: Alan.Geller@microsoft.com 
 ms.date: 12/11/2017
 ms.topic: article
 ---
 
-# Statements and Other Constructs
+# Statements, Control Flow and other constructs 
+
+This section contains description of statements,work flow elements like loops or conditonals and other kind of constructs.
```
cgranade: "workflow elements" reads a bit awkwardly to me, but I'm not sure what to suggest instead, sorry.

articles/manual/statements.md
```diff
@@ -181,7 +185,20 @@ languages such as C#, except that Q# symbols, once bound, may not be changed;
 
 An immutable binding consists of the keyword `let`, followed by
 a symbol or symbol tuple, an equals sign `=`, an expression to bind the symbol(s) to, and a terminating semicolon.
-The type of the bound symbol(s) is inferred based on the expression on the right hand side.
+The type of the bound symbol(s) is inferred based on the expression on the right hand side. For instance:
+
+```qsharp
+let measurementOperator = [PauliX, PauliZ, PauliZ, PauliX, PauliI];
```
cgranade:
```suggestion
let measurementOperators = [PauliX, PauliZ, PauliZ, PauliX, PauliI];
```

articles/manual/statements.md
```diff
@@ -181,7 +185,20 @@ languages such as C#, except that Q# symbols, once bound, may not be changed;
...
+```
+
+This assigns a particular array of Pauli operators to a variable called `measurementOperator`.
```
cgranade:
```suggestion
This assigns a particular array of Pauli operators to an immutable variable called `measurementOperators`.
```

articles/manual/statements.md
```diff
@@ -596,8 +644,10 @@ fail $"Syndrome {syn} is incorrect";
 
 Note that none of these statements are allowed within the body of a function.
 They are only valid within operations.
+### Allocating Qubits
```
cgranade:
```suggestion

### Allocating Qubits
```

articles/manual/statements.md
```diff
@@ -658,7 +712,41 @@ from qubits that are in use but that are not accessed during the body of the `bo
...
 
-## Expression Evaluation Statements
+
+### Intrinsic Operations
+Once allocated, a qubit can then be passed to functions and operations.
+In some sense, this is all that a Q# program can do with a qubit, as the actions that can be taken are all defined as operations.
+More information of these operations can be found in [Intrinsic Operations and Functions](xref:microsoft.quantum.libraries.standard.prelude) and [Operations and Functions](xref:microsoft.quantum.manual.operations-functions).
+
+Here we present an example of operation implementation:
+First, the single-qubit Pauli operators $X$, $Y$, and $Z$ are represented in Q# by the intrinsic operations `X`, `Y`, and `Z`, each of which has type `(Qubit => Unit is Adj + Ctl)`.
```
cgranade: It might be good to link to the API references for `X`, `Y`, and `Z` here.

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
 ```
+## Measurements ##
+
+Using the `Measure` operation, which is a built in intrinsic non-unitary operation, we can extract classical information from an object of type `Qubit` and assign a classical value as a result, which has a reserved type `Result`, indicating that the result is no longer a quantum state. 
+The input to `Measure` is a Pauli axis on the Bloch sphere, represented by an object of type `Pauli` (i.e., for instance `PauliX`) and an object of type `Qubit`. 
+
+A simple example is the following operation which creates one qubit in the $\ket{0}$ state, then applies a Hadamard gate ``H`` to it and then measures the result in the `PauliZ` basis. 
+
+```qsharp
+operation MeasurementOneQubit () : Result {
```
cgranade:
```suggestion
operation MeasureOneQubit() : Result {
```

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+        let result = M(qubit);
+        // As the qubit is now in an eigenstate of the measurement operator, 
+        // we reset the qubit before releasing it. 
+        if (result == One) { X(qubit); }   
+        // Finally, we return the result of the measurement. 
+        return result;
+    }
+}
+```
+
```
cgranade: It may be good to note that this can be done in a simpler fashion by using the libraries:

```Q#
open Microsoft.Quantum.Measurement;

operation MeasureOneQubit() : Result {
   // The following using block creates a fresh qubit and initializes it 
    // in the |0〉 state.
    using (qubit = Qubit()) {
        // We apply a Hadamard operation H to the state, thereby creating the 
        // state 1/sqrt(2)(|0〉+|1〉). 
        H(qubit); 
        // Now we measure the qubit in Z-basis and return the result.
        return MResetZ(qubit);
    }
}

File: articles/manual/statements.md [hunk 21]
```diff
@@ -672,3 +760,130 @@ X(q);
...
+A slightly more complicated example is given by the following operation which returns the Boolean value `true` if all qubits in a register of type `Qubit[]` are in the state zero, when measured in a specified Pauli basis and `false` otherwise.
```
cgranade:
```suggestion
A slightly more complicated example is given by the following operation which returns the Boolean value `true` if all qubits in a register of type `Qubit[]` are in the state zero when measured in a specified Pauli basis, and `false` otherwise. 
```

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+        // Finally, we return the result of the measurement. 
+        return result;
+    }
+}
+```
+
+A slightly more complicated example is given by the following operation which returns the Boolean value `true` if all qubits in a register of type `Qubit[]` are in the state zero, when measured in a specified Pauli basis and `false` otherwise. 
+
+```qsharp
+operation AllMeasurementsZero (qs : Qubit[], pauli : Pauli) : Bool {
```
cgranade: In keeping with the style guide, it would be good if `qs` was the second argument.

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+until (expression)
+fixup {
+    statementBlock2
+}
+```
+where `statementBlock1` and `statementBlock2` are zero or more Q# statements, and `expression` any valid expression that evaluates to a value of type `Bool`. 
+In a typical use case, the following circuit implements a rotation around an irrational axis of $(I + 2i Z)/\sqrt{5}$ on the Bloch sphere. This is accomplished by using a known RUS pattern: 
+
+```qsharp
+operation RUScircuit (qubit : Qubit) : Unit {
```
cgranade:
```suggestion
operation RUSCircuit (qubit : Qubit) : Unit {
```

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+
+    using(ancillas = Qubit[2]) {
```
cgranade:
```suggestion
    using (auxillaryQubits = Qubit[2]) {
```

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+```qsharp
+repeat {
+    statementBlock1 
+}
+until (expression)
+fixup {
+    statementBlock2
+}
+```
+where `statementBlock1` and `statementBlock2` are zero or more Q# statements, and `expression` any valid expression that evaluates to a value of type `Bool`.
```
cgranade:
```suggestion
where `statementBlock1` and `statementBlock2` are zero or more Q# statements, and where `expression` is any valid expression that evaluates to a value of type `Bool`. 
```

articles/manual/statements.md
```diff
@@ -672,3 +760,130 @@ X(q);
...
+```
+
+This example shows the use of a mutable variable `finished` which is in scope of the entire repeat-until-fixup loop and which gets initialized before the loop and updated in the fixup step.
+
+Finally, we show an example of a RUS pattern to prepare a quantum state $\frac{1}{\sqrt{3}}\left(\sqrt{2}\ket{0}+\ket{1}\right)$, starting from the $\ket{+}$ state. See also the [unit testing sample provided with the standard library](https://github.com/Microsoft/Quantum/blob/master/Samples/src/UnitTesting/RepeatUntilSuccessCircuits.qs): 
+
+```qsharp
+operation RepeatUntilSuccessStatePreparation( target : Qubit ) : Unit {
+
+    using( ancilla = Qubit() ) {
```
cgranade:
```suggestion
    using (auxillaryQubit = Qubit()) {
```

articles/manual/index.md
geduardo: I solved the issue in cd720b60367aedac7bba5f431034ff0881c41c2e , now it shows up correctly. 
And yes, I agree, the video is not completely appropriate, it's quite general about Quantum Computing and barely mentions anything related specifically with Q#, but I couldn't find anything better.

articles/manual/index.md
geduardo: Mmm, yes, you're right, it can scare the reader (and maybe "Pre-required knowledge" is a bit rough). However, I think there should be somewhere a small list of essential concepts that one should know. For example, Quantum Katas are great but they start assuming the reader is familiar with concepts such as complex numbers, quantum states, qubits, superposition... which a lot of readers might not be familiar with. Stepping into tutorials that explain things in terms of concepts the reader potentially doesn't know could also scare some users. Books don't suffer from this problem, but most of the first contacts with Q# will likely happen through the most accessible tutorials available.

Although I'm not the best person to judge this because I was already familiar with these concepts when I started to go through the katas. Maybe we should consult someone who started the tutorials with a non-physics background to ask what is the optimal way of present it.

articles/manual/index.md
geduardo: Ok, I like the idea of not splitting learning Q# and learning Quantum Computing. Although I think this whole subsection should be restructured to a bigger section (see #487)

articles/manual/basics.md
geduardo: Ok, I removed the .NET reference in 1b962068a4e558bb9f0642c812f73217e3ae814d

articles/manual/statements.md
geduardo: How about "control flow structures"?

articles/manual/statements.md
geduardo: Ok. Done in 6d27732ae2a8bf026a01f135d6fd4e7891328e2c

articles/manual/statements.md
geduardo: Cool, added in 20241a3d46fc38cf0db98e5aabffbddea3601ac0 and beb92d72a29afc12a2b8177985e0c00526a9be9a

articles/manual/statements.md
geduardo: Done in d84cc0443e3ede4d3a4a48d96a843054728a74d4

articles/manual/index.md
geduardo: I don't fully get the idea you have in mind here. Check 0d24276bcc258abdaa682dd973a725279954fb4b to see if now it's better.

articles/manual/basics.md
geduardo: Done!

KittyYeungQ:
Leave this PR open until #663 resolves

gillenhaalb:
Closing, subsumed in #663 (and #743 )

PR #297, opened 2018-10-27
Update quantum-concepts-4-Qubit.md

fadi-wassaf:
Expanded further on derivation/generation of the single-qubit rotation unitary matrices.

PR #493, opened 2019-08-23
Re-implement #297 against latest master

natke:


cgranade:
Looks like the documentation preview formatted the rotation matrices correctly:

image

articles/concepts/the-qubit.md
```diff
@@ -43,7 +42,7 @@ A final important property of measurement is that it does not necessarily damage
 Qubits may also be pictured in $3$D using the [*Bloch sphere*](https://en.wikipedia.org/wiki/Bloch_sphere) representation.  The Bloch sphere gives a way of describing a single-qubit quantum state (which is a two-dimensional complex vector) as a three-dimensional real-valued vector.  This is important because it allows us to visualize single-qubit states and thereby develop reasoning that can be invaluable in understanding multi-qubit states (where sadly the Bloch sphere representation breaks down).  The Bloch sphere can be visualized as follows:
 
 <!--- ![](.\media\bloch.svg){ width=50% } --->
-![Bloch sphere](~/media/concepts_bloch.png)
+![Bloch sphere](../media/concepts_bloch.png)
```
cgranade: My understanding was that absolute paths like `~` are favorable to relative paths like `..` to ensure stability as content is reorganized. It looks like this PR as well as #508 and #509 change to use relative paths to media, though; is that the best practice going forward?

natke: Yes, you're right. I was getting a warning from the build system when I created this PR which I thought was related to the path. I will change it back and verify that it still works.

Issue #487, opened 2019-08-14
Suggestion: New "Learning resources" section


geduardo:
As a recent learner of Q# I found that there are a lot of learning resources but they are sparsed or not explained in the main website. For example, I've been reviewing the [main site](https://docs.microsoft.com/en-us/quantum/?view=qsharp-preview) for one month and I've found just now the two fantastic compilations of resources in https://qsharp.community/ and https://github.com/ebraminio/awesome-qsharp/ by reading issue #466 instead of from the main site. 

Most new users will start using and learning Q# through the material presented in the main site, but there is not a complete compilation of learning resources on the main site like the ones cited above (or even an _easy to find_ link to them).

There's the section "For more information", but it seems more oriented to provide useful references for advanced topics of Quantum Computing rather than learning material for Q#.

I wrote a small subsection [Learning materials](https://review.docs.microsoft.com/es-es/quantum/manual/?view=qsharp-preview&branch=qmanual#learning-materials) at the beginning of Q# Manual, but still is incomplete and it just will contribute to create more sparsity by adding a new different compilation in another random place.

I think we should create a section (or maybe a subsection on Getting started or Q# manual (?)) on the main site as an **official and structured compilation of learning resources**, explaining what the resources are for and how to use them depending on the level of expertise of the reader. This will help the reader to have a fixed point of reference for obtaining material. It will also help us by needing to keep up to date one single compilation that we can refer to in other sections, blogs... instead of having to update manually each compilation every time a new resource comes out.

Please, tell me what you think @cgranade, @bettinaheim, @bromeg. As I'm learning Q# from almost zero I could take care of and improve it as I learn and follow the different tutorials.

crazy4pi314:
I am glad that https://qsharp.community/ is helpful! We are always looking for help collecting and maintaining resources, as well as blog posts about learning Q# and tools in that ecosystem.

cgranade:
I think that's a great idea, and one that fits nicely in with how Microsoft docs tend to list resources more generally. For instance, Azure Functions provides a number of such resources in a separate TOC section at the bottom:

![image](https://user-images.githubusercontent.com/31516/64457028-71913100-d0a6-11e9-8d25-bc9742642b33.png)

Tagging @natke for further discussion as well.

PR #505, opened 2019-08-29
Updated TOC as a first step towards standard docs.microsoft.com struc…

v-stgr:
…ture. No articles under the TOC have been updated yet.
This TOC may need to go through a few iterations for this PR to get the formatting right and to integrate in reviewer feedback, so please feel free to comment on this PR.
After this TOC has been integrated into the published quantum docs. The next phase will be to update some of the articles under the updated TOC including the initial landing page, overview page, etc. Typically at one page per PR.
Thanks.

### DIFF
--- a/articles/TOC.yml
+++ b/articles/TOC.yml
@@ -22,6 +22,11 @@
   - name: Quantum computing concepts
     topicHref: concepts/index.md
     href: concepts/toc.yml
+  - name: Q# libraries
+    href: libraries/toc.yml
+  - name: Q# programming language
+    topicHref: language/index.md
+    href: language/toc.yml
 - name: How-to guides
   items:
   - name: Managing quantum machines and drivers
@@ -34,16 +39,19 @@
   items:
   - name: Glossary
     href: Glossary.md
-  - name: Q# libraries
-    href: libraries/toc.yml
-  - name: Q# programming language
-    topicHref: language/index.md
-    href: language/toc.yml
+  - name: Q# library
+    href: ../api/qsharp/toc.yml
+  - name: C# components
+    items:
+    - name: Trace simulator 
+      href: https://docs.microsoft.com/dotnet/api/microsoft.quantum.simulation.simulators.qctracesimulators
+    - name: Quantum chemistry
+      href: https://docs.microsoft.com/dotnet/api/microsoft.quantum.chemistry
 - name: Resources
   items:
   - name: Quantum blog
     href: https://cloudblogs.microsoft.com/quantum/
-  - name: Q# blog
+  - name: Q# development blog
     href: https://devblogs.microsoft.com/qsharp/
   - name: For more information
     href: for-more-info.md
--- a/articles/TOC.yml
+++ b/articles/TOC.yml
@@ -1,33 +1,54 @@
-- name: Welcome
+- name: Quantum Development Kit Preview Documentation
   href: index.md
-- name: Getting Started
-  topicHref: install-guide/index.md
-  href: install-guide/toc.yml
-- name: Quantum computing concepts
-  topicHref: concepts/index.md
-  href: concepts/toc.yml
-- name: Quickstart - your first quantum program
-  href: quickstart.md
-- name: Managing quantum machines and drivers
-  topicHref: machines/index.md
-  href: machines/toc.yml
-- name: Quantum development techniques
-  topicHref: techniques/index.md
-  href: techniques/toc.yml
-- name: Q# libraries
-  href: libraries/toc.yml
-- name: Q# programming language
-  topicHref: language/index.md
-  href: language/toc.yml
-- name: Contributing to the Quantum Development Kit
-  href: contributing/toc.yml
-  topicHref: contributing/index.md
-- name: Glossary
-  href: Glossary.md
-- name: For more information
-  href: for-more-info.md
-- name: Release Notes
-  topicHref: relnotes/index.md
-  href: relnotes/toc.yml
-- name: License
-  href: license.md
+- name: Overview
+  items:
+  - name: Getting started
+    href: install-guide/index.md
+  - name: Contributing to the Quantum Development Kit
+    href: contributing/toc.yml
+    topicHref: contributing/index.md
+- name: Quickstarts
+  items:
+  - name: C#
+    href: install-guide/csharp.md
+  - name: Create a Bell State in Q#
+    href: quickstart.md
+  - name: Jupyter Notebooks
+    href: install-guide/jupyter.md
+  - name: Python
+    href: install-guide/python.md
+- name: Concepts
+  items:
+  - name: Quantum computing concepts
+    topicHref: concepts/index.md
+    href: concepts/toc.yml
+- name: How-to guides
+  items:
+  - name: Managing quantum machines and drivers
+    topicHref: machines/index.md
+    href: machines/toc.yml
+  - name: Quantum development techniques
+    topicHref: techniques/index.md
+    href: techniques/toc.yml
+- name: Reference
+  items:
+  - name: Glossary
+    href: Glossary.md
+  - name: Q# libraries
+    href: libraries/toc.yml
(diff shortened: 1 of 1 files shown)


articles/TOC.yml
```diff
@@ -1,33 +1,54 @@
...
+    topicHref: machines/index.md
+    href: machines/toc.yml
+  - name: Quantum development techniques
+    topicHref: techniques/index.md
+    href: techniques/toc.yml
+- name: Reference
+  items:
+  - name: Glossary
+    href: Glossary.md
+  - name: Q# libraries
```
cgranade: At the moment, the libraries section isn't really written as a reference. Does that cause any problems here?

articles/TOC.yml
```diff
@@ -1,33 +1,54 @@
...
+    href: install-guide/csharp.md
+  - name: Create a Bell State in Q#
+    href: quickstart.md
+  - name: Jupyter Notebooks
+    href: install-guide/jupyter.md
+  - name: Python
+    href: install-guide/python.md
+- name: Concepts
+  items:
+  - name: Quantum computing concepts
```
cgranade: The "quantum computing concepts" section is disconnected enough from Q# and the Quantum Development Kit that I worry a little bit about it being the _only_ item listed under "Concepts." Some of the content elsewhere about the Q# language and libraries might could come forward to here?

articles/TOC.yml
```diff
@@ -1,33 +1,54 @@
...
+    href: techniques/toc.yml
+- name: Reference
+  items:
+  - name: Glossary
+    href: Glossary.md
+  - name: Q# libraries
+    href: libraries/toc.yml
+  - name: Q# programming language
+    topicHref: language/index.md
+    href: language/toc.yml
```
cgranade: The Q#, .NET, and Python APIs are all also reference items, should they be linked to from here?

articles/TOC.yml
```diff
@@ -1,33 +1,54 @@
...
+- name: Resources
+  items:
+  - name: Quantum blog
+    href: https://cloudblogs.microsoft.com/quantum/
+  - name: Q# blog
```
cgranade:
```suggestion
  - name: Q# development blog
```

articles/TOC.yml
```diff
@@ -1,33 +1,54 @@
...
+- name: Overview
+  items:
+  - name: Getting started
+    href: install-guide/index.md
+  - name: Contributing to the Quantum Development Kit
+    href: contributing/toc.yml
+    topicHref: contributing/index.md
+- name: Quickstarts
+  items:
+  - name: C#
```
cgranade: I like separating out quickstarts like this, but at the moment it feels like a mix of install instructions (C#, Jupyter / IQ#, and Python) and of quickstarts that help users write their first Q# program without reference to the choice of development environment or host language. For example, I don't see "Create a Bell state in Q#" as an alternative task to "Python," but as something I might want to do once I have Python set up to work with Q#.

v-stgr: Hi Chris - Natalie asked that I submit an initial rework of the TOC without restructuring any articles within the existing site.  So that is why the "Getting started" material is under quickstarts, even though it is mostly prerequisite material.  I have reworked the Bell State in Q# quickstart offline to have a prerequisite section that contains the necessary steps that are currently in Getting started.  Typically Azure programming articles have a prerequisite section in case a reader navigates to it directly from a Google search, etc.  So essentially what you see is an interim step, not the final vision for the TOC.

articles/TOC.yml
v-stgr: I will change this.

articles/TOC.yml
v-stgr: The Q# Library Reference and C# Components Reference are currently linked from the TOC.yml file as they exist outside of docs.microsoft.com. I also would like to move them under References and need to discuss how to move them more with Carolyn and Natalie.  I'm not sure how they are referenced for the TOC at the moment.  Good call in noticing this.  I didn't see any Python specific library docs for the Quantum SDK. Is there a link you can share?  Thanks.

articles/TOC.yml
v-stgr: I'm not sure. Perhaps other team members will chime in with their thoughts after the holiday weekend.  Thanks for pointing this out.

articles/TOC.yml
v-stgr: I think it looks close enough for a first draft of the TOC reorg.  Anyone else?

articles/TOC.yml
geduardo: I agree with @cgranade, probably the only item listed under concepts would be "quantum computing concepts". I didn't find anything in Q# Language that could have its subsection there. I think some parts of the "Q# standard libraries" could fit in "quantum computing concepts". 
As an idea maybe we could create a section about "Computer programming concepts" with general concepts about programming languages. For example, I come from a pure physics background and it's not rare that I need to look up in Google some terminology of the documentation related to CS concepts. I think a short section with the main things (what's a compiler, namespaces, etc) in the context of Q# could be helpful. But I don't know to what extent it is worth it.

geduardo:
"Quantum development techniques" and "Q# programming language" are being merged into a bigger single section with the tentative title "Q# manual" (#486). Maybe you should know this for the organization of the TOC.
And what do you think about replacing the "How-to guides" section with a bigger section "Learning resources" (or something similar) in which the "How-to guides" are included as well as other tutorials and learning resources (more info about my idea of this section in #487)?

articles/TOC.yml
v-stgr: I looked at the Q# libraries and Q# programming concepts sections again. I think they are conceptual enough to move over to the Concepts section, and modify as necessary in subsequent doc revisions. Meanwhile, the C# library reference and C# components reference should be moved up from the bottom of the TOC to the Reference section.

v-stgr:
Based on your feedback, I changed the Q# blog title, moved Q# libraries and Q# programming language to Concepts, and added Q# library and C# components to reference sections.

The Q# library reference and C# components reference at the bottom of the TOC have to be updated in a separate file as part of the build system.

Please have a look and let me know how it looks.  Is this acceptable for a first draft TOC change?

v-stgr:
@bromeg

articles/TOC.yml
cgranade: Makes sense. We may want to re-examine the concepts in the Bell state quickstart as well, then, to be clear about the learning goals of that page. At the moment, we've treated Bell state preparation as a stand-in for a simple task one could perform as a part of a larger Q# application; it's nice from the standpoint that also demonstrates an example of entanglement, but we have definitely seen issues where people have confused "Bell state preparation" with "Bell tests" (aka CHSH games).

Promoting Bell state preparation to the title of the quickstart puts the specifics of the example ahead of the more general learning outcomes we had previously (see below), such that I agree with this being an interim step that will need some more iteration.

image

articles/TOC.yml
cgranade: Makes sense, that sounds like a good path forward!

bromeg:
Looks good to me--I agree with Chris' point about needing to rework the Quickstart section so that the different paths/intent of each article is clear. For some reason in the preview I see the library and component references appearing twice--once in "References" and once after the rest of the nav pane. I couldn't figure out where this was happening based on the TOC.yml file, but we definitely don't want to duplicate those sections.

v-stgr:
Thanks, Megan. The additional components references at the bottom are defined elsewhere, but I'm not sure where otherwise I would have taken them out of the preview (Carolyn and I dicussed these at one point). Perhaps Natalie or someone else knows where these are defined in the TOC?

natke:
This is a great start to refactoring the TOC and gives us a taste of where we want to head. We need to flesh out the getting started experience, and add some tutorials before we publish this TOC change, so that we can have maximum impact. I will close for now. We can revisit these changes in a couple of weeks when we have re-authored the getting started experience and developed one or two tutorials.

PR #508, opened 2019-09-03
Changed png /media file links

v-stgr:
...from ~ to .. to correct build errors.


v-stgr:
@bromeg

v-stgr:
@cgrande can you please review and accept if the change looks okay?

cgranade:
My understanding from discussion with @natke on #493  was that `~` is preferred to `..` to avoid breaking links when reorganizing content. Should we be moving to relative paths instead?

v-stgr:
Oh, I thought you had switched to .. instead of ~ because of build errors.

Issue #359, opened 2019-02-21
Maintain and update LaTeX notation


cgranade:
There's a lot of small LaTeX rendering issues here and there caused by the DocFX -> MathJax pipeline that would be good to resolve. In particular, we've set up macros like `\ket`, `\bra`, and `\id` to help typeset different notation, such that relying more heavily on these macros can help make sure that things render consistently.

- [x] Replace `\left|\psi\right\rangle`-style notation with `\ket{\psi}` (same with bra)
- [x] Replace `\mathbb{1}` with `\id`
- [ ] Fix improperly escaped linebreaks

gillenhaalb:
Opened a branch for this. Any chance you could point out an example of the improper escape?

The few spots where I didn't update the \xangle's was where the bra-ket is an inner product. The current \braket macro is defined as simply <x>, so it appears we need to define a new one for <x|y>. I tried for a bit but wasn't sure how to make the 2 argument macro in the JSON syntax.

gillenhaalb:
@cgranade where do we stand with the linebreaks? Good to close the issue?

cgranade:
I think so, yes. We haven't run into examples of lines running together recently, and can easily open more focused issues should we do so in the future.

PR #510, opened 2019-09-04
gibeck/latex-notation-update

gillenhaalb:
Comments in the "Issue" area (#359). Only things missing are ~~the inner product macro and~~ improper escape fixes. 
Edited: saw that we're meant to throw the "|" in manually for braket inner prod, updated all accordingly in the last commit below.


api/qsharp/microsoft.quantum.diagnostics.assertqubitisinstatewithintolerance.yml
```diff
@@ -26,7 +26,7 @@ remarks: |-
 
   The tolerance is
   the $L\_{\infty}$ distance between 3 dimensional real vector (x₂,x₃,x₄) defined by
-  $\langle\psi|\psi\rangle = x\_1 I + x\_2 X + x\_3 Y + x\_4 Z$ and real vector (y₂,y₃,y₄) defined by
```
cgranade: These files are automatically overwritten from API documentation comments in the source code itself. These changes are useful, such that I want to make sure we capture them; would you be willing to apply your changes for the API reference to the Q# sources in Microsoft/QuantumLibraries instead? Thanks!

gillenhaalb: Absolutely! I'll let you know if I have any trouble branching off that (in case I'm not listed as a contributor to that repo as well). While I'm at it, would you like me to change the expressions currently in unicode(?) style to latex as well? 
e.g. in this specific paragraph alone:
Capture
I'm not sure how widespread it is throughout the docs, but it's a smidge unsightly/cumbersome.

cgranade: Thanks! As for Unicode vs TeX notation, our general strategy has been to use Unicode only where LaTeX notation isn't supported (e.g.: in samples and VS / VSCode–based katas). The `/// # Summary` section of API documentation comments is used in the widest variety of contexts, such that math notation should be avoided there as best as possible (`/// # Description` and `/// # Remarks` are preferred sections for mathematical content).

gillenhaalb: That makes sense, thanks for the clarification. New point of confusion though: I can't find anything in Microsoft/QuantumLibraries that needs updating. Even searching for "\langle," "\rangle," and "\mathbb" merely returned a few spots that I hadn't updated in the first place because they're not technically bras or kets, they just happen to use the angle brackets (e.g. "<ZZI, IZZ>" referring to group generators). 

Not sure if I'm missing something, but I checked out and searched the whole repo in Visual Studio. Just let me know!

cgranade: Ah, my apologies, I had forgotten that AssertQubitIsInStateWithinTolerance is one of the several functions and operations that gets pulled in from the runtime and not the standard libraries. This YAML file is generated from the docs at https://github.com/microsoft/qsharp-runtime/blob/ebb4b2a9de0110ab45a6e38b38cfbd664759483f/src/Simulation/Intrinsic/Diagnostics/AssertQubit.qs#L104. Thanks!

articles/libraries/standard/applications.md
```diff
@@ -182,29 +182,29 @@ Having seen how the quantum Fourier transform and phase estimation work (see [Qu
 Given two integers $a$ and $N$, where $a<N$, the goal of period finding, also called order finding, is to find the _order_ $r$ of $a$ modulo $N$, where $r$ is defined to be the least positive integer such that $a^r \equiv 1 \text{ mod } N$.  
 
 To find the order using a quantum computer, we can use the phase estimation algorithm applied to the following unitary operator $U_a$:
-$$ U_a|x\rangle \equiv |(ax)\text{ mod }N\rangle .$$
+$$ U_a\ket{x} \equiv \ket{(ax)\text{ mod }N} .$$
```
cgranade: One more catch, sorry, `\text{mod}` doesn't space the modulus operator correctly; I think that `\bmod` is preferred, but I haven't checked if we support that yet, sorry. In general, text commands in LaTeX should only be used for text, not operators and gates. Also happy to set that aside for future work, fixing incorrect use of `\text` definitely isn't the highest of priorities...

Issue #507, opened 2019-09-03
Incorrect Jordan Wigner transform


christopherkang:
[Enter feedback here]
In the transformation from the creation operator to its Pauli equivalent, the summation doesn't actually equal the given matrix. Similarly, the annihilation operator doesn't equal it's Pauli equivalent. The actual formulas are used later on in the text, but the first mention is incorrect: 

<img width="493" alt="Screen Shot 2019-09-03 at 10 53 16 AM" src="https://user-images.githubusercontent.com/42285636/64196596-0d1d6a00-ce39-11e9-89fa-b04946cddfd7.png">

(also, if possible, the spacing should be fixed)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: de4f0bcf-e335-8173-6761-d3440f84db82
* Version Independent ID: 98d49390-a026-46ce-2886-19969fe24224
* Content: [Jordan-Wigner Representation](https://docs.microsoft.com/en-us/quantum/libraries/chemistry/concepts/jordan-wigner?view=qsharp-preview)
* Content Source: [articles/libraries/chemistry/concepts/jordan-wigner.md](https://github.com/MicrosoftDocs/quantum-docs-pr/blob/live/articles/libraries/chemistry/concepts/jordan-wigner.md)
* Product: **quantum**
* GitHub Login: @nathanwiebe2
* Microsoft Alias: **nawiebe@microsoft.com**

cgranade:
Thanks for pointing this out! It looks like somehow 𝑎ⱼ⁺ and 𝑎ⱼ got swapped with each other (ironically). The spacing issue seems to be because `\\\` was used instead of `\\\\` as is required for MathJax to see a `\\` through the Markdown processor.

cgranade:
Closing, as this has been addressed with #513. Please feel free to reopen if needed, or to file another issue with follow-ups. Thanks for the report!

christopherkang:
#513 is still broken - switching the matrix forms was incorrect. Here are the final formulas

![image](https://user-images.githubusercontent.com/42285636/64460270-d2bd0280-d0ae-11e9-9f07-ca724dd43015.png)

Additionally, the line before describing the annihilation / creation operators has an error. Those formulas are also provided.

christopherkang:
Current: 
[...]

PR #513, opened 2019-09-05
Fixing issue #507

geduardo:
Switching the matrices and trying to fix the spacing

### DIFF
--- a/articles/libraries/chemistry/concepts/jordan-wigner.md
+++ b/articles/libraries/chemistry/concepts/jordan-wigner.md
@@ -23,8 +23,8 @@ This means that qubits can naturally store the occupation of a given spin orbita
 We then have that $a^\dagger_j \ket{0}_j = \ket{1}_j$ and $a^\dagger_j \ket{1}_j = 0$.
 It is easy to verify that
 \begin{align}
-a^\dagger_j &= \begin{bmatrix}0 & 0 \\\ 1 &0 \end{bmatrix}=\frac{X_j + iY_j}{2}, \nonumber\\\
-a_j &= \begin{bmatrix}0 & 1 \\\ 0 &0 \end{bmatrix}=\frac{X_j - iY_j}{2},
+a^\dagger_j &= \begin{bmatrix}0 & 1 \\\ 0 &0 \end{bmatrix}=\frac{X_j + iY_j}{2}, \nonumber\\\\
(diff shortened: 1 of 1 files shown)


geduardo:
I don't know if the equation for the full construction should be also reformatted. Having to scroll to see the full expression seems a bit awkward to me when it could be split into different lines.

cgranade:
@geduardo: Thanks for fixing this! I agree, having to scroll isn't great; I suspect that there's a similar `\\\` versus `\\\\` issue on that line as well.

geduardo:
Exactly, all of them were `///`. I've checked in the rest of the pages of this section to look for more versions of this error and didn't find anything. So I think it's now ready to merge.

articles/libraries/chemistry/concepts/jordan-wigner.md
```diff
@@ -45,10 +45,10 @@ Thus, by interspersing $Z$ operators into the construction of the operator, we c
...
 \begin{align}
-a^\dagger_1 &= \left(\frac{X-iY}{2}\right)\otimes 1 \otimes 1 \otimes 1 \otimes \cdots \otimes 1,\\\
-a^\dagger_2 &= Z\otimes\left(\frac{X-iY}{2}\right)\otimes 1\otimes 1 \otimes \cdots \otimes 1,\\\
-a^\dagger_3 &= Z\otimes Z\otimes \left(\frac{X-iY}{2}\right)\otimes 1 \otimes \cdots \otimes 1,\\\
-&\vdots\\\
+a^\dagger_1 &= \left(\frac{X-iY}{2}\right)\otimes 1 \otimes 1 \otimes 1 \otimes \cdots \otimes 1,\\\\
+a^\dagger_2 &= Z\otimes\left(\frac{X-iY}{2}\right)\otimes 1\otimes 1 \otimes \cdots \otimes 1,\\\\
+a^\dagger_3 &= Z\otimes Z\otimes \left(\frac{X-iY}{2}\right)\otimes 1 \otimes \cdots \otimes 1,\\\\
+&\vdots\\\\
 a^\dagger_N &= Z\otimes Z\otimes Z\otimes Z \otimes \cdots \otimes Z\otimes \left(\frac{X-iY}{2}\right). \tag{★}\label{eq:JW}
```
cgranade: It looks like `\tag{★}` is breaking the spacing here, plus we never use the `eq:JW` reference anywhere.
```suggestion
a^\dagger_N &= Z\otimes Z\otimes Z\otimes Z \otimes \cdots \otimes Z\otimes \left(\frac{X-iY}{2}\right). \label{eq:JW}
```

PR #525, opened 2019-09-20
Fix OPS build warnings

natke:
- Remove private docset configs (these do not exist)
- Temporarily remove attribute example in the style guide (see #524)
- Update xrefs
- Fix xref_query_tags in the API docset, so that UIDs in the articles docset are resolved

PR #526, opened 2019-09-23
Adjust xref query tags and remove unused docsets

fenxu:
- Add correct xref query tags to fix uid not resolve issue
- Remove unused docset config from configuration file


natke:
Fixed in #525

PR #587, opened 2019-10-31
Adding Bloch sphere to QRNG

geduardo:


articles/quickstarts/qrng.md
```diff
@@ -18,11 +18,33 @@ A simple example of a quantum algorithm written in Q# is a quantum random number
...
+
+Let's visualize with the Bloch sphere how the state of the qubit transforms when we measure it. For example, suppose that we have the state depicted by the red arrow above and we measure it to be **1** (although the odds are low it can happen as well). In the Bloch sphere this process looks like this:
+
+<img src="./Measurement.svg" width="450">
+
+Measurements are not the only way to modify the state of a qubit, we can also apply operations to change the state of the qubit. As any state is represented by a point on the sphere, one operation that transforms the state of the qubit into another can also be understod as a rotation of the arrow along the sphere. For example, the most famous operation that we can apply to a qubit is the `H` operation (H stands for Hadamard, the mathematician who studied it). The `H` operation does the following: takes a qubit in the state **0** and transform it to a balanced state in which the odds of obtaining **0** and **1** are both 50%. In the Bloch sphere:
```
bromeg: I think we should re-use the term superposition here. I'm not sure that people will intuit that the balanced state is a qubit in superposition.

articles/quickstarts/qrng.md
```diff
@@ -18,11 +18,33 @@ A simple example of a quantum algorithm written in Q# is a quantum random number
...
+
+There's no unique way for a qubit to be in superposition, there are infinite ways to exist in superposition. The probabilities of obtaining either **0** or **1** when measuring the qubit depends on the specific superposition. For example, we can have a qubit in superposition in which the probability of obtaining **1** is very high, let's say a 90%. Therefore the probability of obtaining **0** is low (10%). Or maybe we have a qubit in a different superposition in which the probabilities to obtain **0** and **1** when measuring are the same (50%). As there are infinite ways for the qubit to be in a superposition, a clever way to express the state of a qubit is the *Bolch sphere*.
```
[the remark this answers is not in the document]
geduardo: It is common that layman people associate "superposition" with a single state and think that qubits can be in three possibilities: **0**, **1** and **both at the same time**.
But the reality is more nuanced since not all superposition yield the same probabilities of obtaining **0** and **1**. I wanted to convey this. I'll try to condense it to make it clearer.

articles/quickstarts/qrng.md
geduardo: How about "balanced superposition"?

articles/quickstarts/qrng.md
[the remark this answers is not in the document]
cgranade: Agreed, @geduardo, it's really common to see readers that confuse qubits with what are effectively classical trits.

articles/quickstarts/qrng.md
```diff
@@ -18,11 +18,33 @@ A simple example of a quantum algorithm written in Q# is a quantum random number
...
+
+<img src="./Bloch.svg" width="175">
```
cgranade: It may be worth checking that this renders right in other browsers (e.g. Safari, non-Chromium Edge, Firefox), as SVG rendering can sometimes vary between browsers.

articles/quickstarts/qrng.md
```diff
@@ -18,11 +18,33 @@ A simple example of a quantum algorithm written in Q# is a quantum random number
...
+
+In the Bloch sphere the north pole represents the classical value **0** and the south pole represents the classical value **1**. Any superposition can be represented by a point in the sphere. This point is typically represented by an arrow starting at the center of the sphere (also called vector). When the closer the end of the arrow to a pole, the higher the probability the qubit collapses into the classical value assigned to that pole when measured. For example, the qubit state represented by the red arrow above has a higher probability of giving the value **0** if we measure it.
```
cgranade:
```suggestion
In the Bloch sphere the north pole represents the classical value **0** and the south pole represents the classical value **1**. Any superposition can be represented by a point on the sphere. This point is typically represented by an arrow starting at the center of the sphere (also called vector). When the closer the end of the arrow to a pole, the higher the probability the qubit collapses into the classical value assigned to that pole when measured. For example, the qubit state represented by the red arrow above has a higher probability of giving the value **0** if we measure it.
```

articles/quickstarts/qrng.md
cgranade: If I understand right, "balanced" here is to indicate that it's not just any superposition, but the uniform superposition $(\ket{0} + \ket{1}) / \sqrt{2}$ that is being discussed?

geduardo: Exactly, couldn't think of a better way to express it without introducing kets. But since it seems you both got slightly confused by it maybe is better to suppress it.

articles/quickstarts/qrng.md
geduardo: In Chrome and old Edge works fine. We can check in other browsers (although it is a bit of a pain to validate MS log-in for each browser xD)

bromeg:
Closing this pull request since we merged the shorter version to master already.

PR #588, opened 2019-10-31
Adding Bloch sphere to QRNG (Short)

geduardo:
This is the PR for the shortened version of the Bloch sphere visualization.


articles/quickstarts/qrng.md
```diff
@@ -48,3 +49,19 @@ Using the `H` operation, we are able to put our `Qubit` in superposition. To mea
 By putting our `Qubit` in superposition and measuring it, our result will be a different value each time the code is invoked. 
 
 When a `Qubit` is de-allocated it must be explicitly set back to the `Zero` state, otherwise the simulator will report a runtime error. An easy way to achieve this is invoking `Reset`.
+
+### Visualizating the code with the Bloch sphere
+
+In the Bloch sphere the north pole represents the classical value **0** and the south pole represents the classical value **1**. Any superposition can be represented by a point in the sphere (represented by an arrow). When the closer the end of the arrow to a pole, the higher the probability the qubit collapses into the classical value assigned to that pole when measured. For example, the qubit state represented by the red arrow below has a higher probability of giving the value **0** if we measure it.
```
cgranade:
```suggestion
In the Bloch sphere the north pole represents the classical value **0** and the south pole represents the classical value **1**. Any superposition can be represented by a point on the sphere (represented by an arrow). When the closer the end of the arrow to a pole, the higher the probability the qubit collapses into the classical value assigned to that pole when measured. For example, the qubit state represented by the red arrow below has a higher probability of giving the value **0** if we measure it.
```

articles/quickstarts/qrng.md
```diff
@@ -48,3 +49,19 @@ Using the `H` operation, we are able to put our `Qubit` in superposition. To mea
...
+
+We can use this representation to visualize what the code is doing:
+
+* First we start with a qubit initalizated in the state **0** and apply `H` to create a superposition in which the probabilities for **0** and **1** are the same.
+
+<img src="./H.svg" width="450">
+
+* Then we measure the qubit and save the output:
+
+<img src="./Measurement2.svg" width="450">
```
bromeg:
```suggestion
<img src="./Measurement2.svg" width="450">

Since we have an equal probability of measuring 0 or 1, we have no way of knowing before we measure which state the `Qubit` will collapse to. Each time we run the algorithm we might get a different result based on the randomness of whether the `Qubit` collapses to 0 or 1. 
```

geduardo:
Done @bromeg, check it out. We can expand that post Ignite to implement the host program to create the integers.

cgranade:
Looks like there's a warning with respect to your author metadata, @bromeg?

articles/quickstarts/qrng.md
```diff
@@ -23,6 +23,7 @@ A simple example of a quantum algorithm written in Q# is a quantum random number
 As mentioned in our [What is Quantum Computing?](xref:microsoft.quantum.overview.what) article, a qubit is a unit of quantum information that can be in superposition. When measured, a qubit can only be either 0 or 1, however during execution, the state of the qubit represents the probability of reading either a 0 or a 1 during measurement. This probabilistic state is known as superposition. We can use this probability to generate random numbers.
 
 ### Q# operation code
+Let's see how we can implement the concepts showed above to create a quantum random number generator in Q#:
 
 1. Replace the contents of the Operation.qs file with the following code:
```
cgranade: I recognize that this isn't part of the changes you've made with this PR, but the source code at lines 32--42 seem like they have some indenting issues?

articles/quickstarts/qrng.md
```diff
@@ -48,3 +49,21 @@ Using the `H` operation, we are able to put our `Qubit` in superposition. To mea
...
+
+We can use this representation to visualize what the code is doing:
+
+* First we start with a qubit initalizated in the state **0** and apply `H` to create a superposition in which the probabilities for **0** and **1** are the same.
+
+<img src="./H.svg" width="450">
+
+* Then we measure the qubit and save the output:
+
+<img src="./Measurement2.svg" width="450">
```
cgranade: This image looks a bit off in Edge Dev for me:
image

articles/quickstarts/qrng.md
```diff
@@ -48,3 +49,21 @@ Using the `H` operation, we are able to put our `Qubit` in superposition. To mea
...
+
+Since the outcome obtained by the measurement is completely random we have obtained a random bit. We can call this function several times to create integers. For example, if we call the function three times to obtain three random bits, we can build random 3-bit numbers (this is, a random number between 1 and 8).
```
cgranade:
```suggestion
Since the outcome obtained by the measurement is completely random we have obtained a random bit. We can call this function several times to create integers. For example, if we call the function three times to obtain three random bits, we can build random 3-bit numbers (that is, a random number between 0 and 7).
```

articles/quickstarts/qrng.md
bromeg: oh yeah, Measurement is kind of wonky looking

articles/quickstarts/qrng.md
geduardo: I guess that my latent Matlab programmer took the control there.

PR #607, opened 2019-11-04
Editorial updates to overview documents.

natke:


articles/TOC.yml
```diff
@@ -6,6 +6,8 @@
     href: welcome.md
   - name: Five questions about quantum computing
     items:
+    - name: Overview
```
cjgronlund:
```suggestion
 
```

articles/TOC.yml
```diff
@@ -6,6 +6,8 @@
...
+      href: overview/index.md
```
cjgronlund:
```suggestion

```

cjgronlund: PMs requested we remove this from TOC. Let's do it for now.

articles/overview/quantum-computers.md
```diff
@@ -12,7 +12,9 @@ uid: microsoft.quantum.overview.computers
 
 What we can do with a quantum computer that can't be done with a classical one?
```
cjgronlund:
```suggestion
What can we do with a quantum computer that can't be done with a classical one?
```

articles/TOC.yml
natke: Sure

PR #643, opened 2019-12-09
Fix LaTeX rendering and feedback link.

cgranade:
This PR moves LaTeX macros from global to file metadata to fix microsoft/Quantum#271, and redirects feedback for Q# API references to the libraries repo that's used to generate those references.

### DIFF
--- a/api/docfx.json
+++ b/api/docfx.json
@@ -39,25 +39,31 @@
     "overwrite": [],
     "externalReference": [],
     "globalMetadata": {
-      "latex_macros": {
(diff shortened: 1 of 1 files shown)


cgranade:
Looks like this needs a full rebuild to check...

bromeg:
@cgranade looks like there are merge conflicts, but would be good to get this into the push to live

cgranade:
I think this has since been addressed by @natke's work, actually, such that we can close this PR rather than merging it in.

PR #694, opened 2020-02-06
slight adjustment to the landing page

KittyYeungQ:


articles/index.yml
```diff
@@ -66,10 +66,10 @@ landingContent:
             url: machines/index.md
       - linkListType: quickstart
         links:
-          - text: Build a quantum random number generator
-            url: quickstarts/qrng.md
           - text: Quantum basics with Q#
             url: quickstart.md
+          - text: Build a quantum random number generator
```
geduardo: I'm not sure if we should change the order here. The QRNG tutorial was created because the Bell example that appears in _Quantum basics with Q#_ was a bit complicated for a "quantum hello world" tutorial.

KittyYeungQ: In that case, the order in TOC needs to change for consistency.

KittyYeungQ: And we need to change the title "Quantum basics with Q#" as it is misleading. Any suggestions of an alternative?

geduardo: Yup, I agree, how about: "Exploring entanglement with Q#"?

KittyYeungQ: That's a good title.

PR #707, opened 2020-02-19
Update microsoft.quantum.arrays.filtered.yml

KittyYeungQ:


api/qsharp/microsoft.quantum.arrays.filtered.yml
```diff
@@ -40,4 +40,16 @@ output:
 typeParameters:
 - uid: "'T"
   summary: The type of `array` elements.
+examples: |-
+  The following code demostrates the "Filtered" function. A predicate is defined using a function called "GreaterTahnI" from the Microsoft.Quantum.Logical namespace:
```
Bradben: typo in the function name GreaterThanI

cgranade:
I'd strongly advise against changing these YAML files directly, as any changes will get overwritten the next time they are automatically generated from the relevant Q# libraries. In this case, I'd suggest making sure that all changes originate at microsoft/QuantumLibraries#231, so that they persist when the documentation is regenerated.

tcNickolas:
This pull request (and others that target API docs directly) are supposed to serve as a preview of the changes, since we cannot generate such previews from pull requests on QuantumLibraries directly. Maybe we should introduce a tag on this repository "not for merge"?

cgranade:
> This pull request (and others that target API docs directly) are supposed to serve as a preview of the changes, since we cannot generate such previews from pull requests on QuantumLibraries directly.

Makes perfect sense, only left that comment as I noticed that some changes were _only_ reflected in this PR and not microsoft/QuantumLibraries#231.

> Maybe we should introduce a tag on this repository "not for merge"?

Marking PRs as draft can help, as can leaving a review with "request changes." If we rely only on a label, that won't prevent someone accidentally merging unless we add a new CI check (possibly with GitHub Actions) to look for that label. That could definitely be worth doing, but until then, I'd suggest reviewing with request changes is an effective guard against merging.

PR #739, opened 2020-03-04
add index to API ref to fix 404, add indexes to API toc

Bradben:


api/index.md
```diff
@@ -9,36 +9,6 @@ ms.topic: landing-page
...
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
\ No newline at end of file
+In addition to the operations and functions of the Q# *standard library*, there are also .NET and Python libraries available for Q#.
```
cgranade: The API reference also covers the three Q# libraries built on top of the standard library: chemistry, numerics, and machine learning.

api/qsharp/index.md
```diff
@@ -0,0 +1,14 @@
...
+author: bradben
+ms.author: bradben
+ms.date: 03/04/2020
+ms.topic: landing-page
+---
+
+
+# Q# libraries #
+
+Q# is supported by a range of different useful operations, functions, and user-defined types that comprise the Q# *standard library*.
```
cgranade: Same comment here as well. Mentioning the other three libraries as well as the standard library would be helpful, I think.

api/qsharp/toc.yml
```diff
@@ -1,3 +1,5 @@
+- uid: microsoft.quantum.qsharplibintro
```
cgranade: This TOC.yml file is automatically generated and will be overwritten by our current publication pipeline.

cgranade:
Thanks for adding this, @bradben. I've marked as request changes since you've suggested valuable changes to automatically generated files, such that I want to make sure we don't accidentally lose those changes the next time we publish API docs.

api/qsharp/toc.yml
Bradben: Thanks, @cgranade.  Is the entire qsharp/api/qsharp folder overwritten or are the files just updated? I ask because we need to keep an index.md file in that folder to prevent a 404 error. I can workaround not using the toc.yml file.

Bradben: Reverted the changes on api/qsharp/toc.yml

cgranade: > Thanks, @cgranade. Is the entire qsharp/api/qsharp folder overwritten or are the files just updated? I ask because we need to keep an index.md file in that folder to prevent a 404 error. I can workaround not using the toc.yml file.

The process is fairly manual at this point, in that we copy the files emitted by the build by hand into qsharp/api/qsharp. Those files include toc.yml such that changes there are definitely overwritten, but changes to index.md can be kept if we're very careful when updating API docs.

PR #758, opened 2020-03-18
update redirects for names changes and deletions

Bradben:
Moved (and name change):
articles/Glossary.md -> articles/concepts/glossary.md

Name change:
install-guide/python.md -> pyinstall.md
install-guide/csharp.md -> csinstall.md
install-guide/jupyter.md -> qjupyter.md

Deleted:
install-guide/vs-2017.md -> /install-guide/csinstall.md
install-guide/vs-code.md -> /install-guide/csinstall.md
install-guide/command-line.md -> /install-guide/csinstall.md
articles/concepts/software-stack.md -> /concepts

### DIFF
--- a/.openpublishing.redirection.json
+++ b/.openpublishing.redirection.json
@@ -53,7 +53,7 @@
         },
         {
             "source_path": "articles/quantum-concepts-9-SoftwareStack.md",
-            "redirect_url": "concepts/software-stack",
+            "redirect_url": "concepts/",
             "redirect_document_id": true
         },
         {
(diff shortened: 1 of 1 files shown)


natke:
Hi @Bradben, do you think there is a good reason to change the filenames? Could we change them back and avoid the redirects?

Bradben:
> Hi @Bradben, do you think there is a good reason to change the filenames? Could we change them back and avoid the redirects?

I did a grep of the repo and the filenames are only used in TOC.yml. We can change the filenames back and change TOC.yml to match. I don't see any reason to keep the redirects.

Bradben:
After a second cup of coffee, I realized we could leave everything as is and just not add the redirects.

PR #802, opened 2020-05-01
Update lineranges for microsoft/quantum#346.

cgranade:


cgranade:
Looks like the line numbers all worked out. @rmshaffer, would you be willing to re-approve? Thanks!

PR #804, opened 2020-05-01
Prepare standalone docs for publication

cgranade:
Marked as draft until #803 completes.

PR #805, opened 2020-05-01
Publish all docs for 0.11.2004 release.

cgranade:
Draft until #804 completes.


cgranade:
I'm not sure why this is raising warnings about code samples not being found, but I suspect it's because the docs build ran an incremental build. Let me try and force that.

cgranade:
The only warnings are about the new IQ# docs, which we've confirmed still render correctly. While we should fix those warnings, I think we're good to defer that for now. Thanks!

PR #810, opened 2020-05-04
TOC update: Content model compliance

cjgronlund:
TOC updated to comply at the top-level nodes to the Content & Learning doc model. More changes to article arrangement to come.


articles/TOC.yml
```diff
@@ -46,15 +44,23 @@
...
-- name: QDK Release Notes
-  topicHref: relnotes/index.md
-  href: relnotes/toc.yml
+- name: Concepts 
+  items: 
+  - name: Quantum computing concepts
+    href: concepts/toc.yml
+  - name: Q# libraries
+    href: libraries/toc.yml
+- name: Reference
```
cgranade, line 53: Looking at the rendered TOC, this creates an empty "Reference" node that may be easily confused with the "API Reference" node lower down in the TOC. If possible, it would be good to fix that before going live.

image

cjgronlund: Good call. We'll remove it.

articles/TOC.yml
```diff
@@ -46,15 +44,23 @@
...
-  topicHref: relnotes/index.md
-  href: relnotes/toc.yml
+- name: Concepts 
+  items: 
+  - name: Quantum computing concepts
+    href: concepts/toc.yml
+  - name: Q# libraries
+    href: libraries/toc.yml
+- name: Reference 
+  items:
```
cjgronlund, line 54:
```suggestion
```

articles/TOC.yml
cjgronlund, line 53:
```suggestion
```

cjgronlund:
Close PR and create new one.

PR #813, opened 2020-05-05
Small fixes (TOC and Update page)

gillenhaalb:
Hides "How to create a Q# project" from TOC (it's outdated to the point of causing confusion, and will be subsumed by coming pages anyway)

On Update page, changed `netcoreapp3.0` to `netcoreapp3.1`

PR #816, opened 2020-05-06
Update changes made before May 6 2020

KittyYeungQ:


geduardo:
Apparently there's a conflict with the TOC file

KittyYeungQ:
Open a new PR

KittyYeungQ:
@cgranade we are observing something strange here. Not sure why the build failed. The TOC file was already merged from master to live but it's showing up again here. The only change after that was #813 we'd like to merge to live.

cgranade:
Very odd indeed. There should never be conflicts between live and master by construction; I wonder if something went wrong with a previous merge to live, and just didn't happen to affect anything until modifying the same file again? Easy enough to fix, though; I can remove branch protection on master for a moment, merge live into master, push it back, then master will be strictly ahead of live.

(What I wouldn't give to be able to set merge policies per-branch instead of per-repo...)

cgranade:
I'll get on that as soon as I can, apologies for the delay.

KittyYeungQ:
Thank you very much @cgranade

cgranade:
It should be fixed, please let me know if anything else happens.

PR #955, opened 2020-07-07
Update JoinTocPlugin configuration

yufeih:
This JoinTocPlugin config affects 2 docsets, so extract it to the root node. This will unblock docfx v3 migration for this repo.

@TianqiZhang

### DIFF
--- a/.openpublishing.publish.config.json
+++ b/.openpublishing.publish.config.json
@@ -23,15 +23,7 @@
         "docset_prebuild": [
           "_dependentPackages/CommonPlugins/tools/JoinTOC.ps1"
         ]
-      },
-      "JoinTOCPlugin": [
-        {
-          "ReferenceTOC": "api/TOC.yml",
-          "ConceptualTOC": "articles/TOC.yml",
-          "ReferenceTOCUrl": "/qsharp/api/toc.json",
-          "ConceptualTOCUrl": "/quantum/toc.json"
-        }
-      ]
+      }
     },
     {
       "docset_name": "quantum-docs",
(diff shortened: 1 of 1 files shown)


geduardo:
Hi @yufeih @TianqiZhang,  we need to do something that I've been told that it is possible to do but it's a bit convoluted. I post this here since I see you are modifying the _JoinTOCPlugin_ 

We have currently two repositories:

- **quantum-docs-pr:** it's the current repository that holds the source code for the current live quantum documentation. This repository currently contains two different docsets. A reference docset under the folder /articles and an automatically generated docset for the API under /api. These two docsets are joined in the TOC using the ´JoinTOCPlugin´. 

- **quantum-docs-private:** this repository is a private copy of the previous one but we are adding the content for the documentation for Azure quantum. We have been told that we need the documentation to be under an ´/azure´ URL since it is an Azure service. But we don't want to move all the existing documentation for the Quantum Development Kit (´/quantum´ )  to a new Azure docset since apparently it will be very problematic in terms of search equity, broken links, etc. So what we are doing is to create a new azure docset under the folder /azure-quantum, to store all the documentation related to the Azure platform. So this repository will have three different docsets: /azure-quantum (conceptual), /articles (conceptual) and /api (reference).

 
We would like to display all the docsets under a single TOC in the following way:

- Azure Quantum
- Quantum Development Kit 
- API reference

But the documentation for the JoinTOCPlugin only offers information for the fusion of only two docsets. Do you know how should we proceed?

TianqiZhang:
@geduardo I'm afraid that JoinTOC only supports two TOCs right now. @rloutlaw for more information on the future of JoinTOC/FusionTOC.

KittyYeungQ:
Thanks @TianqiZhang @rloutlaw @cillroy Would be helpful to know if there is any plan to support more than two TOCs and if so when that may start.

yufeih:
Hi @geduardo @KittyYeungQ , would you help review and merge this PR?

PR #958, opened 2020-07-08
Replace with "opaque box"

marktab:
Per https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/b/black-box  replacing with "opaque box"


KittyYeungQ:
Consulted team implementing site-wide changes. This term is not listed to be changed.

PR #1047, opened 2020-09-10
Fix typo for issue #1046

### LINKED ISSUE #1046: Empty Tuple Code Sample in Entanglement Tutorial
While explaining the `Unit `keyword, it made a reference to an empty tuple in Python which was followed by this code sample:
`(Tuple[()])`. However, this is incorrect, as an empty tuple can be initialized by either `tuple()` or `tuple([])`, but _not_ `Tuple[()]`

This is a relatively minor issue, but considering the nature of the documentation and rigour of the material, I feel like this is a valid albeit inconsequential issue.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f0f20167-a5ab-1b18-feca-17f4ffd7d1e3
* Version Independent ID: ef2aa8d4-af46-57ee-567e-12f23eeb0822
* Content: [Explore entanglement with Q# - Microsoft Quantum](https://docs.microsoft.com/en-us/quantum/tutorials/explore-entanglement)
* Content Source: [articles/tutorials/explore-entanglement.md](https://github.com/MicrosoftDocs/quantum-docs-pr/blob/live/articles/tutorials/explore-entanglement.md)
* Product: **quantum**
* GitHub Login: @geduardo
* Microsoft Alias: **v-edsanc@microsoft.com**

geduardo:
Fixes #1046

### DIFF
--- a/articles/tutorials/explore-entanglement.md
+++ b/articles/tutorials/explore-entanglement.md
@@ -152,7 +152,7 @@ The arguments to an operation are specified as a tuple, within parentheses.
 The return type of the operation is specified after a colon. In this case, the
 `SetQubitState` operation has no return, so it is marked as returning `Unit`. This is the
 Q# equivalent of `unit` in F#, which is roughly analogous to `void` in C#, and
-an empty tuple (`Tuple[()]`) in Python.
+an empty tuple (`tuple[()]`) in Python.
 
 You have used two quantum operations in your first Q# operation:
 
--- a/articles/tutorials/explore-entanglement.md
+++ b/articles/tutorials/explore-entanglement.md
@@ -152,7 +152,7 @@ The arguments to an operation are specified as a tuple, within parentheses.
 The return type of the operation is specified after a colon. In this case, the
 `SetQubitState` operation has no return, so it is marked as returning `Unit`. This is the
 Q# equivalent of `unit` in F#, which is roughly analogous to `void` in C#, and
-an empty tuple (`tuple[()]`) in Python.
+an empty tuple in Python (`()`, represented by the type hint `Tuple[()]`).
 
 You have used two quantum operations in your first Q# operation:
 


articles/tutorials/explore-entanglement.md
```diff
@@ -152,7 +152,7 @@ The arguments to an operation are specified as a tuple, within parentheses.
 The return type of the operation is specified after a colon. In this case, the
 `SetQubitState` operation has no return, so it is marked as returning `Unit`. This is the
 Q# equivalent of `unit` in F#, which is roughly analogous to `void` in C#, and
-an empty tuple (`Tuple[()]`) in Python.
+an empty tuple (`tuple[()]`) in Python.
```
cgranade, line 155: The Python value `()` can also be written as `tuple([])`, but not as `tuple[()]`. On the other hand, parameters to type hints in Python are denoted by `[]` rather than `()`, so the type of `tuple([])` is `Tuple[()]`.

cgranade (changes requested):

articles/tutorials/explore-entanglement.md
geduardo: So I guess we can close this PR, right?

cgranade: I don't think so, it's a genuine confusion we can address! I went with `Tuple[()]` as that's most directly analogous to `unit` in F# and `void` in C#. The Python value `()` or `tuple([])` is most directly analogous to F#'s `()` rather than `unit`. That said, as you point out, type hints aren't something widely known in Python yet, so it may be helpful to take that feedback as impetuous to clarify:

```suggestion
an empty tuple in Python (`()`, represented by the type hint `Tuple[()]`).
```

cgranade (approved):

PR #1053, opened 2020-09-19
Fix some typos

wsgac:


articles/concepts/index.md
```diff
@@ -13,7 +13,7 @@ no-loc: ['Q#', '$$v']
...
+There are many ways to understand why quantum mechanics is hard to simulate.  Perhaps the simplest is to see that quantum theory can be interpreted as saying that matter, at a quantum level, is in a host of different possible configurations (known as *states*) at the same time.  Unlike classical probability theory, these many configurations of the quantum state, which can be potentially observed, may interfere with each other like waves in a tide pool.  This interference prevents the use of statistical sampling to obtain the quantum state configurations.  Rather, we have to track *every possible* configuration a quantum system could be in if we want to understand the quantum evolution.
```
KittyYeungQ, line 16:
```suggestion
There are many ways to understand why quantum mechanics is hard to simulate.  Perhaps the simplest is to see that quantum theory can be interpreted as saying that matter, at a quantum level, is in a host with different possible configurations (known as *states*).  Unlike classical probability theory, these many configurations of the quantum state, which can be potentially observed, may interfere with each other like waves in a tide pool.  This interference prevents the use of statistical sampling to obtain the quantum state configurations.  Rather, we have to track *every possible* configuration a quantum system could be in if we want to understand the quantum evolution.  
```

Bradben: ..."can be interpreted as saying that matter, at a quantum level, is in a host of different possible configurations (known as *states*) at the same time. "

It might be clearer to say ..."can be interpreted as saying that matter, at a quantum level, is in a multitude of possible configurations (known as *states*) at the same time.

Bradben (approved):
One comment, otherwise approved - thanks!

articles/concepts/index.md
KittyYeungQ: I think we should also remove "at the same time".

Bradben: Okay - ..."can be interpreted as saying that matter, at a quantum level, is in a multitude of possible configurations, known as *states*."

KittyYeungQ (approved):

PR #1059, opened 2020-09-25
edit pass on standard library topics, pt. 2

Bradben:


articles/user-guide/libraries/standard/data-structures.md
```diff
@@ -174,52 +176,59 @@ $$
...
+## Dynamical generator modeling in Q\#
+
+Generators of time-evolution describe how states evolve through time. For example, the dynamics of a quantum state $\ket{\psi}$ is governed by the Schrödinger equation
 $$
 \begin{align}
     i\frac{d \ket{\psi(t)}}{d t} & = H \ket{\psi(t)},
 \end{align}
 $$
-with a Hermitian matrix $H$, known as the Hamiltonian, as the generator of motion. Given an initial state $\ket{\psi(0)}$ at time $t=0$, the formal solution to this equation at time $t$ may be, in principle, written
+with a Hermitian matrix $H$, known as the [Hamiltonian](xref:microsoft.quantum.chemistry.concepts.quantumdynamics), as the generator of motion. Given an initial state $\ket{\psi(0)}$ at time $t=0$, the formal solution to this equation at time $t$ may be, in principle, written
```
cgranade, line 215: This link does to the chemistry definition of a Hamiltonian, which is an important special case of the more general definition (a linear operator describing the energy of a system). If we need a link for this, I'd recommend making a new conceptual page for it, instead of communicating that chemistry is the only context in which Hamiltonian evolution occurs.

articles/user-guide/libraries/standard/data-structures.md
```diff
@@ -174,52 +176,59 @@ $$
...
 \end{align}
 $$
 
-where time-evolution by each term alone is easy to implement on a quantum computer. For instance, if $H_j$ is a Pauli $X_1X_2$ operator acting on the 1st and 2nd elements of the qubit register `qubits`, time-evolution by it for any time $t$ may be implemented simply by calling the operation `Exp([PauliX,PauliX], t, qubits[1..2])`, which has signature `((Pauli[], Double, Qubit[]) => Unit is Adj + Ctl)`. As discussed later in Hamiltonian Simulation, one solution then is to approximate time-evolution by $H$ with a sequence of simpler operations
+where time-evolution by each term alone is easy to implement on a quantum computer. For example, if $H_j$ is a Pauli $X_1X_2$ operator acting on the first and second elements of the qubit register `qubits`, time-evolution by it for any time $t$ may be implemented simply by calling the operation `Exp([PauliX,PauliX], t, qubits[1..2])`, which has signature `((Pauli[], Double, Qubit[]) => Unit is Adj + Ctl)`. As discussed in [Simulating Hamiltonian Dynamics](xref:microsoft.quantum.chemistry.concepts.simulationalgorithms), one solution is to approximate time-evolution by $H$ with a sequence of simpler operations
```
cgranade, line 231: Similarly here, I'd suggest generalizing the scope of the linked content to not convey that simulation is limited to chemistry. While that's a critically important special case, there's also very important applications outside of chemistry.

articles/user-guide/libraries/standard/data-structures.md
```diff
@@ -250,61 +259,61 @@ The `GeneratorSystem` user-defined type is a classical description of the Hamilt
...
+### Pauli operator generators
 
 A concrete and useful example of generators are Hamiltonians that are a sum of Pauli operators, each possibly with a different coefficient.
 $$
 \begin{align}
     H & = \sum^{d-1}_{j=0} a_j H_j,
 \end{align}
 $$
-where each $\hat H_j$ is now drawn from the Pauli group. For such systems, we provide the `PauliEvolutionSet()` of type `EvolutionSet` that defines a convention for how an element of the Pauli group and a coefficient may be identified by a `GeneratorIndex`, which has the following signature.
+where each $\hat H_j$ is now drawn from the Pauli group. For such systems, Q# provides the `PauliEvolutionSet()` of type `EvolutionSet` that defines a convention for how an element of the Pauli group and a coefficient may be identified by a `GeneratorIndex`, which has the following signature:
```
cgranade, line 280: Technically, the Q# standard library provides this, rather than the language.

articles/user-guide/libraries/standard/diagnostics.md
```diff
@@ -30,20 +31,20 @@ Message($"About to rotate by an angle of {angle}...");
...
+## Facts and assertions
 
 As discussed in [Testing and Debugging](xref:microsoft.quantum.guide.testingdebugging), a function or operation with signature `Unit -> Unit` or `Unit => Unit`, respectively, can be marked as a *unit test*.
 Each unit test generally consists of a small quantum program, along with one or more conditions that check the correctness of that program.
 These conditions can come in the form of either _facts_, which check the values of their inputs, or _assertions_, which check the states of one or more qubits passed as input.
 
-For example, `EqualityFactI(1 + 1, 2, "1 + 1 != 2")` represents the mathematical fact that $1 + 1 = 2$, while `AssertQubit(One, qubit)` represents the condition that measuring `qubit` will return a `One` with certainty.
-In the former case, we can check the correctness of the condition given only its values, while in the latter, we must know something about the state of the qubit in order to evaluate the assertion.
+For example, `EqualityFactI(1 + 1, 2, "1 + 1 != 2")` represents the mathematical fact that $1 + 1 = 2$, while `AssertQubit(One, qubit)` represents the condition that measuring `qubit` returns a `One` with certainty.
```
cgranade, line 46: The distinction between "returns" and "will return" is important here, such that the former is not actually correct. Assertions don't check the return value of a measurement operation (that would be a fact), but check the counterfactual condition that an _if_ an operation is run, then the return value will be correct, without actually running the given measurement operation.

Put differently, `EqualityFactR(M(qubit), One, "")` actually performs the measurement such that there is a concrete `Result` value to compare, while `AssertQubit(One, qubit)` does not perform a measurement at all but ensures that if a measurement were to be performed, it would always return `One` at that point.

articles/user-guide/libraries/standard/diagnostics.md
```diff
@@ -88,12 +89,12 @@ That is,
 \begin{align}
     \ket{\psi} = \ket{\psi'} \text{ if and only if } \braket{\psi | \psi'} = 1.
 \end{align}
-Using the primitive operations defined in the prelude, we can directly perform a measurement that returns `Zero` if $\ket{\psi}$ is an eigenstate of one of the Pauli operators.
+Using the primitive operations defined in the prelude, you can directly perform a measurement that returns `Zero` if $\ket{\psi}$ is an eigenstate of one of the Pauli operators.
```
cgranade, line 92: This looks like another place where we can update language; "primitive" → "intrinsic", for example, and getting rid of "prelude."

articles/user-guide/libraries/standard/diagnostics.md
```diff
@@ -88,12 +89,12 @@ That is,
...
 
 
-The operation <xref:microsoft.quantum.diagnostics.assertqubit> provides a particularly useful shorthand to do so in the case that we wish to test the assertion $\ket{\psi} = \ket{0}$.
-This is common, for instance, when we have uncomputed to return ancilla qubits to $\ket{0}$ before releasing them.
-Asserting against $\ket{0}$ is also useful when we wish to assert that two state preparation `P` and `Q` operations both prepare the same state, and when `Q` supports `Adjoint`.
+The operation <xref:microsoft.quantum.diagnostics.assertqubit> provides a particularly useful shorthand to do so in the case that you want to test the assertion $\ket{\psi} = \ket{0}$.
+This is common, for example, when you have uncomputed to return ancilla qubits to $\ket{0}$ before releasing them.
```
cgranade, line 96: Somehow this use of "ancilla" escaped earlier edits; can we modify to "auxiliary?"

articles/user-guide/libraries/standard/diagnostics.md
```diff
@@ -105,14 +106,14 @@ using (register = Qubit()) {
 }
 ```
 
-More generally, however, we may not have access to assertions about states that do not coincide with eigenstates of Pauli operators.
-For example, $\ket{\psi} = (\ket{0} + e^{i \pi / 8} \ket{1}) / \sqrt{2}$ is not an eigenstate of any Pauli operator, such that we cannot use <xref:microsoft.quantum.diagnostics.assertmeasurementprobability> to uniquely determine that a state $\ket{\psi'}$ is equal to $\ket{\psi}$.
-Instead, we must decompose the assertion $\ket{\psi'} = \ket{\psi}$ into assumptions that can be directly tested using  the primitives supported by our simulator.
+More generally, however, you may not have access to assertions about states that do not coincide with eigenstates of Pauli operators.
+For example, $\ket{\psi} = (\ket{0} + e^{i \pi / 8} \ket{1}) / \sqrt{2}$ is not an eigenstate of any Pauli operator, such that you cannot use <xref:microsoft.quantum.diagnostics.assertmeasurementprobability> to uniquely determine that a state $\ket{\psi'}$ is equal to $\ket{\psi}$.
+Instead, you must decompose the assertion $\ket{\psi'} = \ket{\psi}$ into assumptions that can be directly tested using the primitives supported by your simulator.
```
cgranade, line 111:
```suggestion
Instead, you must decompose the assertion $\ket{\psi'} = \ket{\psi}$ into assumptions that can be directly tested using the intrinsics supported by your simulator.
```

articles/user-guide/libraries/standard/diagnostics.md
```diff
@@ -159,7 +160,7 @@ By contrast, the referenced assertion calls each operation exactly once, such th
 Both of these tests are useful in ensuring the correctness of quantum programs.
 
 
-## Further Reading ##
+## Next steps
 
-- <xref:microsoft.quantum.guide.testingdebugging>
-- <xref:microsoft.quantum.diagnostics>
+- Read more about unit tests, facts and assertions, dump functions, and debugging in Visual Studio in <xref:microsoft.quantum.guide.testingdebugging>.
```
cgranade, line 165: That's a good point; can we broaden `testingdebugging` to not so strongly privilege Visual Studio, given that's no longer the strict majority of our users?

articles/user-guide/libraries/standard/error-correction.md
```diff
@@ -1,36 +1,41 @@
...
+That is, if any of the three bits are flipped, then you can recover the state of the logical bit by taking a majority vote.
+Though classical error correction is a much richer subject that this particular example, the code here points to a possible problem in protecting quantum information.
+Namely, the [no-cloning theorem](xref:microsoft.quantum.concepts.pauli#the-no-cloning-theorem) implies that if you measure each qubit and take a majority vote by analogy to classical code in the example, then you have lost the precise information that you are trying to protect.
+
+> [!TIP]
+> For a background in classical error correction, see [Lint's introduction to coding theory](https://www.springer.com/us/book/9783540641339).
+
+In the quantum setting, the measurement is problematic. However, you can still implement the encoding example.
+It is helpful to do so to see how you can generalize error correction to the quantum case.
 Thus, let $\ket{\overline{0}} = \ket{000} = \ket{0} \otimes \ket{0} \otimes \ket{0}$, and let $\ket{\overline{1}} = \ket{111}$.
```
cgranade, line 36: Should L36 still open with "Thus"?

articles/user-guide/libraries/standard/error-correction.md
```diff
@@ -51,15 +56,15 @@ To see how we can identify that this is the case without measuring the very stat
...
-For example, if we measure $Z_0$, we get a different result for $\ket{\overline{0}}$ and $\ket{\overline{1}}$ in the no-error case, so that collapses the encoded state.
-On the other hand, consider measuring $Z_0 Z_1$, the parity of the first two bits in each computational basis state.
-Recall that each measurement of a Pauli operator checks which eigenvalue  the state being measured corresponds to, so for each state $\ket{\psi}$ in the table above, we can compute $Z_0 Z_1 \ket{\psi}$ to see if we get $\pm\ket{\psi}$.
-Note that $Z_0 Z_1 \ket{000} = \ket{000}$ and that $Z_0 Z_1 \ket{111} = \ket{111}$, so that we conclude that this measurement does the same thing to both encoded states.
+To protect the state that you're encoding, you need to be able to distinguish the three errors from each other and the identity $\boldone$ without distinguishing between $\ket{\overline{0}}$ and $\ket{\overline{1}}$.
+For example, if you measure $Z_0$, you get a different result for $\ket{\overline{0}}$ and $\ket{\overline{1}}$ in the no-error case, so that collapses the encoded state.
```
cgranade, line 60: It looks like in this article, $Z_0$ refers to `PauliZ` on the first qubit and `PauliI` elsewhere, but in the above article, that was $Z_1$. Can we perhaps take the opportunity to normalize the two a little bit? Relatedly, it may help to provide a concrete call to the Q# `Measure` operation here.

articles/user-guide/libraries/standard/index.md
```diff
@@ -3,18 +3,31 @@ title: Introduction to Microsoft Q# standard libraries
...
 no-loc: ['Q#', '$$v']
 ---
 
-# Introduction to the Q# Standard Libraries
+# Introduction to the Q# standard libraries
 
 Q# is supported by a range of different useful operations, functions, and user-defined types that comprise the Q# *standard libraries*.
-The [`Microsoft.Quantum.Development.Kit` NuGet package](https://www.nuget.org/packages/microsoft.quantum.development.kit) installed during [installation and validation](xref:microsoft.quantum.install) provides the Q# compiler, and parts of the standard library that are implemented by the target machines.
-The [`Microsoft.Quantum.Standard` package](https://www.nuget.org/packages/microsoft.quantum.standard) provides the portion of the Q# standard libraries that are portable across target machines.
+The [`Microsoft.Quantum.Development.Kit` NuGet package](https://www.nuget.org/packages/microsoft.quantum.development.kit), installed during [installation and validation](xref:microsoft.quantum.install), provides the Q# compiler and the parts of the standard library that are implemented by the [quantum simulators](xref:microsoft.quantum.machines), or target machines.
```
cgranade, line 15: I think this package has since been deprecated, such that users should only depend on `Microsoft.Quantum.Sdk` moving forward. That package pulls in `Microsoft.Quantum.Standard` such that the next line may also not be a user-facing detail any more.

articles/user-guide/libraries/standard/licensing.md
```diff
@@ -4,39 +4,38 @@ title: Open source licensing for the QDK
...
+These specialized functions and operations can be obtained from the [Microsoft/Quantum-NC](https://github.com/microsoft/quantum-nc) repository on GitHub.
 
-There are also many samples that explain and illustrate the use of functions and operations from the standard library and other libraries.
-These samples are licensed under the [MIT license](https://github.com/Microsoft/Quantum/blob/main/LICENSE.txt).
-The samples can be obtained from the [Microsoft/Quantum](https://github.com/Microsoft/Quantum) repository on GitHub.
+There are also many samples available that explain and illustrate the use of functions and operations from the standard library and other libraries.
+These samples are licensed under the [MIT license](https://github.com/Microsoft/Quantum/blob/main/LICENSE.txt) and can be obtained from the [Microsoft/Quantum](https://github.com/Microsoft/Quantum) repository on GitHub.
 
-## Contributing ##
+## Contributing
```
cgranade, line 28: It may help to link to the contribution guide here as well?

articles/user-guide/libraries/standard/prelude.md
```diff
@@ -3,39 +3,37 @@ title: Intrinsic operations and functions in the QDK
...
+Many of the functions and operations defined in this portion of the prelude are in the @"microsoft.quantum.intrinsic" namespace, such that most Q# source files have an `open Microsoft.Quantum.Intrinsic;` directive immediately following the initial namespace declaration.
+The Q# compiler opens the <xref:microsoft.quantum.core> namespace automatically, so that functions such as <xref:microsoft.quantum.core.length> can be used without an `open` statement at all.
 
-Many of the functions and operations defined in this portion of the prelude are in the @"microsoft.quantum.intrinsic" namespace, such that most Q# source files will have an `open Microsoft.Quantum.Intrinsic;` directive immediately following the initial namespace declaration.
-The <xref:microsoft.quantum.core> namespace is automatically opened, so that functions such as <xref:microsoft.quantum.core.length> can be used without an `open` statement at all.
-
-### Common Single-Qubit Unitary Operations ###
+## Common single-qubit unitary operations in Q#
```
cgranade, line 31: Does this need escaped?
```suggestion
## Common single-qubit unitary operations in Q\#
```

articles/user-guide/libraries/standard/prelude.md
```diff
@@ -291,34 +278,34 @@ for (index in 0..Length(qs)-1)
 return rs;
 ```
 
-## Extension Functions and Operations ##
+## Extension functions and operations
```
cgranade, line 281: The concept of extension functions and operations is no longer really used; that exposed an implementation detail about the .NET-based runtime for Q# that has now been abstracted away much better, such that users aren't required to memorize what parts of the Q# library are implemented in pure Q# and which parts are delegated to the runtime.

cgranade (commented):
I didn't do a terrible through review yet since this is still in draft, but I wanted to provide my early feedback. Please let me know if there's anything else it would be helpful for me to review. Thanks!

articles/user-guide/libraries/standard/prelude.md
Bradben: @cgranade - Thanks, this is great feedback. I figured these topics were old enough that significant updates to the content would be needed.  This will help with the next phase.

PR #1060, opened 2020-09-25
add links for new MS Learn quantum modules

Bradben:


articles/tutorials/toc.yml
```diff
@@ -10,3 +10,5 @@
     href: intro-to-katas.md
   - name: Q# code samples
     href: https://docs.microsoft.com/samples/browse/?languages=qsharp
+  - name: MS Learn - Quantum computing fundamentals
```
gillenhaalb, line 13: If I were unfamiliar with Learn (which I'd imagine a decent portion of people will be), I don't know if I'd ever explore this listing in the TOC. Also, while it fits for the title of the learning path, I think in the context of our documentation people might interpret "Quantum computing fundamentals" as implying something which is relatively independent from Q# or the QDK (like the "Concepts" section). 

At the very least, I think including "Learning path" in it might better convey to people what the link leads to. To also include the explicit reference to Learn, maybe something like "Learning path: MS Learn modules"?

Bradben: Good point, thanks.  I poked around other Azure features and most link to their MS Learn content from their main Hub page (which we don't have and most likely won't). However, there is some consistency among Azure features without Hub pages (Azure SQL, Web Apps, and VM) to link from TOC -> Resources -> "Build your skills with MS Learn".  Keeping this consistent wording may work for us.

articles/quickstarts/get-started.md
```diff
@@ -27,13 +27,11 @@ To get up and running with the power of quantum, you can try out our tutorials:
...
 
-### Learning further
-* The [Microsoft Learn modules for quantum computing](https://docs.microsoft.com/learn/browse/?term=quantum) will teach you to master core concepts at your speed and on your schedule. You can learn the basics of how to create quantum programs with the QDK with our [first module](https://docs.microsoft.com/learn/modules/qsharp-create-first-quantum-development-kit/).
-* If you want to dive deeper into Q# programming, check out the [Quantum Katas](https://github.com/Microsoft/QuantumKatas) - a collection of self-paced programming exercises that introduce you to quantum computing via programming exercises in Q#.
-	Many of these katas are also available as Q# Notebooks. 
+## Learning further
+* Microsoft Learn offers free online training for quantum computing. The three-module Learning Path, [Quantum computing foundations](https://docs.microsoft.com/learn/paths/quantum-computing-fundamentals/), introduces the fundamental concepts of quantum computing and quantum algorithms, and gets you started building quantum programs using Q#.
```
geduardo, line 33:
```suggestion
* Microsoft Learn offers free online training for quantum computing. The three-module Learning Path, [Quantum computing foundations](https://docs.microsoft.com/learn/paths/quantum-computing-fundamentals/), introduces the fundamental concepts of quantum computing and quantum algorithms, and gets you started building quantum programs using Q#.
```
Since we are going to add more modules maybe we should avoid mentioning the number of modules to avoid future inconsistencies.

geduardo (approved):

articles/quickstarts/get-started.md
Bradben: Good catch, thanks.

PR #1068, opened 2020-10-05
Merge main to live with new release notes

rmshaffer:

PR #1069, opened 2020-10-05
Address PR feedback from #1068

rmshaffer:


Bradben (commented):
I like these workarounds, thanks!

articles/user-guide/index.md
```diff
@@ -48,7 +48,7 @@ In the different topics of this guide, we detail the core concepts of the Q# lan
 
 ### Quantum Simulators and Resource Estimators
 
-- [Quantum simulators and host applications](xref:microsoft.quantum.machines): An overview of the different simulators available, as well as the general execution model between host programs and target machines.
+- [Quantum simulators and host applications](xref:microsoft.quantum.machines): An overview of the different simulators available, as well as the general interaction model between host programs and target machines.
```
cgranade, line 51: I'm not sure that the term "interaction model" works as a substitute in this case; it looks like in some cases, "execution model" is still used (e.g.: https://docs.microsoft.com/en-us/windows/win32/opengl/execution-model). If that's not in keeping with current guidance, however, I'd suggest reworking this bullet item to avoid the term altogether. Perhaps something like the following?

> An overview of the different simulators available, as well of how host programs and target machines work together to run Q# programs.

articles/user-guide/machines/index.md
```diff
@@ -17,7 +17,7 @@ Quantum simulators are software programs that run on classical computers and act
 
 ## The Quantum Development Kit (QDK) quantum simulators
 
-The quantum simulator is responsible for providing implementations of quantum primitives for an algorithm. This includes primitive operations such as `H`, `CNOT`, and `Measure`, as well as qubit management and tracking. The QDK includes different classes of quantum simulators representing different execution models for the same quantum algorithm. 
+The quantum simulator is responsible for providing implementations of quantum primitives for an algorithm. This includes primitive operations such as `H`, `CNOT`, and `Measure`, as well as qubit management and tracking. The QDK includes different classes of quantum simulators representing different simulation techniques for the same quantum algorithm.
```
cgranade, line 20: I take it this refers to that we have the Toffoli and trace simulators as well as the full-state simulator?

```suggestion
The quantum simulator is responsible for providing implementations of quantum primitives for an algorithm. This includes primitive operations such as `H`, `CNOT`, and `Measure`, as well as qubit management and tracking. The QDK includes different classes of quantum simulators representing different ways of simulating the same quantum algorithm. 
```

articles/user-guide/index.md
rmshaffer: Yeah, that's better!

cgranade (approved):

PR #1040, opened 2020-09-05
Keywords/language reference pages

gillenhaalb:
On hold until migration to Q# language repo is settled (links will all direct to respective pages there).

Addresses the broader discussion in #939. Creates separate pages for each Q# keyword (for searchability). Also some language constructs/symbols.

Content for pages:
- [ ] brief description
- [ ] links to 
    1. User Guide section where it's discussed 
    2. and/or new internal pages from qsharp-language repo (stubs created at PR #1075 )
- [ ] code examples of use (preferably from samples?)

To-do: 
- [ ] group main page contextually? Align with TOC folders, can lend more info than simple alphabetical order (not to mention everyone will likely ctrl+f to find anything they're looking for anyway) 
- [ ] expand EntryPoint and Test API pages because linking to them directly (i.e. add the content that will be on other pages: more explanation, links, examples)

Additional pages to be added:
- `!` unwrap operator
- `< >` type parameterized callables
- `w/ <-` copy-and-update _expressions_ (array elts and named items
- `w/= <-` update-and-reassign _statements_ (used with `set` for named items and array items)
- `? |` conditional expression
- (maybe) new page for _each_ operator? (list [here](https://docs.microsoft.com/en-us/quantum/user-guide/language/expressions#operator-precedence))

PR #1075, opened 2020-10-09
create stub files for qsharp-language topics

Bradben:
(See workflow plan in the wiki for this repo).

First round of stub files for the qsharp-language topics and related TOC updates. 

Next steps:
- First pass on qsharp-language topics to clean up basics: remove H1s, update links to xrefs, and add required metadata.


gillenhaalb:
@Bradben thanks so much for setting this up, I think it'll mesh very nicely with the Language Ref files from #1040.
A few of the points I wanted to bring up are precisely what you have under TBD in the wiki, so I'll just reference those directly:
> 1. Will the existing topic files under Q# language be replaced? (Types in Q# and Expressions)
> 2. Will other topic files under the Use Q# section be merged or replaced?
> 3. How do the Q# language topic files sync with the Language Ref topic files that Gillen is working on?

The folks working on the `qsharp-language` repo can probably speak best to 1 and 2 (@bettinaheim, @samarsha, @cgranade ?). When transferring things into the User Guide format though, one thing that stood out was how easy it would be to accidentally not transfer over some snippet of information; be it crucial or merely tips/words of wisdom. I would recommend then that before any such full replacement goes to master, we have multiple reviewers explicitly compare the two with that in mind.

Regarding 3, it's looking like a mixed bag. Each of those pages/files will be for a single keyword, and then at the bottom of each page (after the explanation/code example) I'm planning on linking to the pages in our documentation where they are discussed in the broader context. So, the page for a given keyword might link to multiple of these `qsharp-language` repo pages and/or others from the User Guide and elsewhere.

geduardo:
Do you know if this is possible @Bradben ?

> Links in the source files should be an xref to the uid of the relative stub topic file. For example,
> 
> https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/ConditionalBranching
> 
> would be
> 
> xref:microsoft.quantum.qsharp.conditional-branching.

Since the qsharp-language repo is not going to be build with DocFX I don't know if we will have a xrefmap that we can add to the the json.config file.

Bradben:
> Do you know if this is possible @Bradben ?
> 
> > Links in the source files should be an xref to the uid of the relative stub topic file. For example,
> > https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/ConditionalBranching
> > would be
> > xref:microsoft.quantum.qsharp.conditional-branching.
> 
> Since the qsharp-language repo is not going to be build with DocFX I don't know if we will have a xrefmap that we can add to the the json.config file.

We're just pulling the raw markdown into our files and building in our repo, so it should work on our side. The question is, are the qsharp-language files just raw source files for us, or is there an expected user experience within the qsharp-language repo? @bettinaheim ?

bettinaheim:
> > Do you know if this is possible @Bradben ?
> > > Links in the source files should be an xref to the uid of the relative stub topic file. For example,
> > > https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/ConditionalBranching
> > > would be
> > > xref:microsoft.quantum.qsharp.conditional-branching.
> > 
> > 
> > Since the qsharp-language repo is not going to be build with DocFX I don't know if we will have a xrefmap that we can add to the the json.config file.
> 
> We're just pulling the raw markdown into our files and building in our repo, so it should work on our side. The question is, are the qsharp-language files just raw source files for us, or is there an expected user experience within the qsharp-language repo? @bettinaheim ?

@Bradben The expectation is just that the markdown content is displayed. There are quite a few links in the files. I am not sure what a good setup there is; I suspect the links will work but clicking on a link that doesn't direct to a section in the same file would direct to the source on the language repo. Is that correct? I think that is fine,  though.

Bradben:
> > > Do you know if this is possible @Bradben ?
> > > > Links in the source files should be an xref to the uid of the relative stub topic file. For example,
> > > > https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/ConditionalBranching
> > > > would be
> > > > xref:microsoft.quantum.qsharp.conditional-branching.
> > > 
> > > 
> > > Since the qsharp-language repo is not going to be build with DocFX I don't know if we will have a xrefmap that we can add to the the json.config file.
> > 
> > 
> > We're just pulling the raw markdown into our files and building in our repo, so it should work on our side. The question is, are the qsharp-language files just raw source files for us, or is there an expected user experience within the qsharp-language repo? @bettinaheim ?
> 
> @Bradben The expectation is just that the markdown content is displayed. There are quite a few links in the files. I am not sure what a good setup there is; I suspect the links will work but clicking on a link that doesn't direct to a section in the same file would direct to the source on the language repo. Is that correct? I think that is fine, though.

@bettinaheim - The links will work when we publish, but the problem is that the user will be in the docs.microsoft.com site and the link will take them out to github, which is not a preferred user experience.  I'm not sure what the answer is yet.

Bradben:
> > > > Do you know if this is possible @Bradben ?
> > > > > Links in the source files should be an xref to the uid of the relative stub topic file. For example,
> > > > > https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/ConditionalBranching
> > > > > would be
> > > > > xref:microsoft.quantum.qsharp.conditional-branching.
> > > > 
> > > > 
> > > > Since the qsharp-language repo is not going to be build with DocFX I don't know if we will have a xrefmap that we can add to the the json.config file.
> > > 
> > > 
> > > We're just pulling the raw markdown into our files and building in our repo, so it should work on our side. The question is, are the qsharp-language files just raw source files for us, or is there an expected user experience within the qsharp-language repo? @bettinaheim ?
> > 
> > 
> > @Bradben The expectation is just that the markdown content is displayed. There are quite a few links in the files. I am not sure what a good setup there is; I suspect the links will work but clicking on a link that doesn't direct to a section in the same file would direct to the source on the language repo. Is that correct? I think that is fine, though.
> 
> @bettinaheim - The links will work when we publish, but the problem is that the user will be in the docs.microsoft.com site and the link will take them out to github, which is not a preferred user experience. I'm not sure what the answer is yet.

@cgranade - any ideas on this? I'm not seeing a way to keep a single source file that has one user experience in GitHub and another user experience on doc.ms.com.  One solution Eduardo and I discussed yesterday would be to build the Q# files in the docs repo monthly, using a script to update the links.

KittyYeungQ:
@bettinaheim can also help unblock.

bettinaheim:
@Bradben @KittyYeungQ @geduardo 
I don't see a better way to keep all links within the docs pages than to publish the language docs manually the way we do the API docs. The script would then need to pull the content from the language repo replace all links to point to the corresponding docs page. @cgranade do you see a way around that, e.g. by defining "aliases/shortcuts" for links in a file somewhere with the idea that this file would be different for the GitHub repo and for the docs page?

cgranade:
> @cgranade - any ideas on this? I'm not seeing a way to keep a single source file that has one user experience in GitHub and another user experience on doc.ms.com. One solution Eduardo and I discussed yesterday would be to build the Q# files in the docs repo monthly, using a script to update the links.
—@Bradben 


> @Bradben @KittyYeungQ @geduardo
> I don't see a better way to keep all links within the docs pages than to publish the language docs using a script that pulls the content from the language repo and replaces all links to point to the corresponding docs page. @cgranade do you see a way around that, e.g. by defining "aliases/shortcuts" for links in a file somewhere with the idea that this file would be different for the GitHub repo and for the docs page? Or maybe it would make sense to have the language repo as submodule of this repo and set up a CI that does these replacements. Any thoughts on that?
—@bettinaheim

If we want to pursue a script, my suggestion would be to have that script live on the qsharp-language repo and to trigger that script from a GitHub Action whenever `main` is updated on qsharp-language. That script could then update a specially named branch (similar to the `★binder` branch on the samples repo), that could then get pulled in by the docs repo, replacing the current dependency on `main`.

As suggested, the main action of such a script could be to replace internal Markdown links with UID links. To enable that, I'd suggest adding DocFX-style metadata headers to the Markdown files in qsharp-language, so that the script could look up the UID that should be inserted for each link. It shouldn't be _too_ bad to write that either as a small .NET program using Markdig (heck, even [a `*.csx` file](https://github.com/filipw/dotnet-script) or C# Jupyter Notebook), or as a Python script using python-frontmatter and https://github.com/Python-Markdown/markdown.

geduardo:
@cgranade @Bradben @bettinaheim  I like the idea of the script and honestly, I don't see another reasonable way to solve this issue. If we go for this option I'm happy to help since I'd love to learn to implement this kind of scripts (I don't know much about GitHub Actions but I'd love to learn).

cgranade:
> @cgranade @Bradben @bettinaheim I like the idea of the script and, honestly, I don't see another reasonable way to solve this issue. If we go for this option I'm happy to help since I'd love to learn to implement this kind of scripts (I don't know much about GitHub Actions but I'd love to learn).

Sounds good! From that perspective, it can be helpful to first make a script that works locally, and then to develop the infrastructure to trigger that script from an Action. Would you want a .NET or Python based approach?

geduardo:
> Sounds good! From that perspective, it can be helpful to first make a script that works locally, and then to develop the infrastructure to trigger that script from an Action. Would you want a .NET or Python-based approach?

I'd go for Python, since I have more expertise there. I can try to create that local version (and I'll probably ask you for help along the way 😄 @cgranade )

PR #1078, opened 2020-10-14
Update toc.yml

KittyYeungQ:
Changed this to match with our new website design. Also we probably shouldn't have called it Q# applications in the first place as it is not consistent with the rest of the TOC titles.


KittyYeungQ:
Resolved on the website.

PR #1079, opened 2020-10-14
Update install-command-line.md

KittyYeungQ:


geduardo:
Mm, this change seems to imply that you need VS Code, when you don't. You can use any IDE/editor and run the applications from the .NET console.

KittyYeungQ:
> Mm, this change seems to imply that you need VS Code, when you don't. You can use any IDE/editor and run the applications from the .NET console.

Can we come up with a title that aligns with the rest of the docs? It is still an instruction on VS/VS Code as opposed to other IDEs. I know we explicitly say this at the beginning but we may need to make it even clearer: 

"The following instructions are using VS and VS Code as an example. However, you can use any environment and run the applications from the .NET console. ...

Follow the instructions at the tab corresponding to your environment.
Q# programs can run on their own, without a driver in a host language like C#, F#, or Python."

geduardo:
> Can we come up with a title that aligns with the rest of the docs? It is still an instruction on VS/VS Code as opposed to other IDEs. I know we explicitly say this at the beginning but we may need to make it even clearer:

> "The following instructions are using VS and VS Code as an example. However, you can use any environment and run the applications from the .NET console. 

I don't know, we also have the instructions for other IDEs with the CLI in the Tabid... Maybe if we migrate to the new Pivots structure we can personalize the #H1 header for each option? I'm not sure about that tho

articles/quickstarts/install-command-line.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Develop with Q# applications
+title: Develop with Q# applications in an IDE
```
Bradben, line 2: The other titles are all "Develop with Q# and ...".  This is consistent but may need tweaking.

articles/quickstarts/install-command-line.md
```diff
@@ -10,46 +10,43 @@ uid: microsoft.quantum.install.standalone
...
 
 - [.NET Core SDK 3.1 or later](https://www.microsoft.com/net/download)
 
 ## Installation
 
-While you can build Q# applications in any IDE, we recommend using Visual Studio Code (VS Code) or Visual Studio IDE for developing your Q# applications locally. For developing in the Cloud via the web browser, we recommend Visual Studio Codespaces. Developing in these environments includes the rich functionality of the QDK extension, which includes warnings, syntax highlighting, project templates, and more. 
+While you can build Q# applications in any IDE, we recommend using Visual Studio Code (VS Code) or Visual Studio IDE for developing your Q# applications locally. For developing in the Cloud via the web browser, we recommend Visual Studio Codespaces. Developing in these environments leverages the rich functionality of the QDK extension, which includes warnings, syntax highlighting, project templates, and more. 
 
-To configure VS Code:
+### To configure VS Code:
```
Bradben, line 25: By making these H3 headers, they stand out equally as install options.

articles/quickstarts/install-command-line.md
```diff
@@ -10,46 +10,43 @@ uid: microsoft.quantum.install.standalone
...
 
-To configure Visual Studio Codespaces:
-
-1. Create an [Azure account](https://azure.microsoft.com/free/).
-2. Create a Codespaces environment. Please follow the [quickstart guide](https://docs.microsoft.com/visualstudio/codespaces/quickstarts/browser). When creating the Codespace, we recommend to enter `microsoft/Quantum` in the "Git Repository" field to load QDK-specific settings.
-3. You can now launch your new environment and start developing in the browser via the [VS Codespaces Cloud IDE](https://online.visualstudio.com/environments). Alternatively, it is possible to use your local installation of VS Code and use Codespaces as a [remote environment](https://docs.microsoft.com/visualstudio/online/how-to/vscode).
-
-
-To install the QDK for another environment, enter the following at the command prompt:
+### To configure the QDK for another environment, enter the following at the command prompt:
```
Bradben, line 35: I moved this up before Codespaces to keep it together with the other local options.

articles/quickstarts/install-command-line.md
```diff
@@ -96,12 +93,6 @@ To run the application:
 
 Verify your installation by creating a Q# `Hello World` application.
 
-1. Install the project templates.
```
Bradben, line 99: To maintain consistency with the VS and VSCode tabs, I took out the step that repeats the install.

articles/quickstarts/install-command-line.md
gillenhaalb: @geduardo mentioned that the old title, "Develop with Q# command-line applications", probably got changed at some point in the restructuring. Thoughts on going back to that @KittyYeungQ ?

geduardo (approved):

PR #1090, opened 2020-10-23
Adapt to new UID format for API reference.

cgranade:
This PR modifies xref links to the API reference to use the new UID format, bringing xref links for Q# in line with C# and Python xrefs by using literal casing rather than lowercasing. Note that this PR depends on new docs being published with the 0.13.2010* release; links modified by this PR will be broken until those API docs are published in their own PR.


geduardo:
Should we wait until the new build is live?

cgranade:
> Should we wait until the new build is live?

At least before merging to `live`, yes, as this PR does depend on the API docs from the new build.

ricardo-espinoza (approved):

PR #1096, opened 2020-10-29
Added initial version of the article Quantum mechanics via QSharp snippets - Part 1

tvarga78:
Based on feedback from IT people, what they'd like to see at the beginning of their quantum journey is a coherent, high-level overview of quantum mechanics, especially with some concrete examples of how superposition is created in the physical world. If they had such an introduction, they would feel much more comfortable to continue.


tvarga78:
No interest, closing it.

PR #1109, opened 2020-11-13
add install icons and reconfigure install table

Bradben:
Added icons for the different dev environments to the install table. I couldn't get a decent layout view with the markdown table so I reconfigured it as an html table - I think it looks much cleaner.


cgranade:
Looks pretty good, though the "Q# Standalone" icon doesn't really work well in dark mode:

image

There also seems to be an odd vertical alignment issue in the second of the header columns?

Bradben:
Hmm, neither does the jupyter icon. Not sure what you mean about the alignment - the visual studio column?

cgranade:
Apologies for poor markup, but it looks like there's an oddly large top margin to some of the text below "OS support:"

Bradben:
Ah... I think I can fix that with a valign tag. I think docfx supports it.

geduardo:
For solving the dark mode issue this might be useful @Bradben (from the README.md in the articles folder): 

Adding images
To have the images rendering properly in dark mode you must avoid transparencies.

For *.jpg files you don't need to do anything since the file format doesn't support transparent elements.
For *.png files you must add a white background or change the value of the alpha channel to 100. The easiest way to do this in Windows is by opening the file in Paint and saving, overwriting the original file.
For *.svg files you must add a white rectangle in the lowest layer. You can do this with Inkscape:
Open the *.svg file.
Select the square maker tool and draw a white rectangle on top of the original figure.
Select the tool "Select and transform objects" by clicking in the dark arrow or pressing F1.
While having the rectangle selected, click in the toolbar element "Lower selection to bottom (End)".
Adjust the rectangle with the mouse or the arrow keys.

Bradben:
Apparently docfx does not support valign (or inline styles), so I let it go with the default top align.  

I also added white background versions of the quantum and juptyer pngs.

geduardo (approved):

PR #1124, opened 2020-11-24
Reference docs for 0.14.2011.120134.

vxfield:
Merge QDK Nov 2020 release Reference Docs into main.


geduardo:
There are many warnings. I guess it is because the new builder isn't adding some required fields in the metadata. Also, some many xrefs give an error, I think it is because they're in lower case.

vxfield:
@cgranade, is this related to the new DocGen?
Can you give me some pointers on how to fix these warning?
And do you think we can release QDK it and just update the docs later? Or this should block the release?

PS: I closed this PR because it's based on the old RC build. The new build 0.14.2011.120240 is still in progress.

vxfield:
@cgranade, let me know if this PR makes sense: https://github.com/microsoft/qsharp-compiler/pull/756

cgranade:
> @cgranade, is this related to the new DocGen?
> Can you give me some pointers on how to fix these warning?
> And do you think we can release QDK it and just update the docs later? Or this should block the release?
> 
> PS: I closed this PR because it's based on the old RC build. The new build 0.14.2011.120240 is still in progress.

@vxfield: I think this shouldn't be blocking, both since these are warnings, and because the required modifications can be made in-place post-release if needed.

PR #1136, opened 2020-11-27
Updating Grover's tutorial (ready for review)

geduardo:
This PR aims to update or complement the existing tutorial for Grover's, by providing a more practical guide and shifting from the database example and bringing it closer to the Grover's Learn module.

Edit: **The content of the article is ready for review.** I still leave it in draft since there are some issues with the xrefs and we need to decide if we want to keep the older tutorial or no.

The tutorial is quite long, so feel free to suggest to drop things that you might consider non-essential.


articles/tutorials/new_search.md
```diff
@@ -0,0 +1,476 @@
...
+Grover's algorithm is one of the most famous algorithms in quantum computing. The problem it solves is often referred to as "searching a database", but it's more accurate to think of it as a "search problem".
+
+> [!NOTE] 
+> This tutorial is intended for people who are already familiar with
+> Grover's algorithm that want to learn how to implement it in Q#. For a more
+> slow paced tutorial we recommend the Microsoft Learn module [Solve graph
+> coloring problems by using Grover's
+> search](https://docs.microsoft.com/learn/modules/solve-graph-coloring-problems-grovers-search/).
+
+Any searching task can be mathematically formulated with an abstract function $f(x)$ that accepts search items $x$. If the item $x$ is a solution for the search task, then $f(x)=1$. If the item $x$ isn't a solution, then $f(x)=0$. The search problem consists on finding any item $x_0$ such that $f(x_0)=1$. This is, an item $x_0$ that is a solution of the search problem.
```
KittyYeungQ, line 25:
```suggestion
Any searching task can be mathematically formulated with an abstract function $f(x)$ that accepts search items $x$. If the item $x$ is a solution for the search task, then $f(x)=1$. If the item $x$ isn't a solution, then $f(x)=0$. The search problem consists of finding any item $x_0$ such that $f(x_0)=1$. This is, an item $x_0$ that is a solution of the search problem.
```

articles/tutorials/new_search.md
```diff
@@ -0,0 +1,476 @@
...
+
+Grover's algorithm is one of the most famous algorithms in quantum computing. The problem it solves is often referred to as "searching a database", but it's more accurate to think of it as a "search problem".
+
+> [!NOTE] 
+> This tutorial is intended for people who are already familiar with
+> Grover's algorithm that want to learn how to implement it in Q#. For a more
+> slow paced tutorial we recommend the Microsoft Learn module [Solve graph
+> coloring problems by using Grover's
+> search](https://docs.microsoft.com/learn/modules/solve-graph-coloring-problems-grovers-search/).
+
```
KittyYeungQ, line 24: I wonder if the MS Learn module is going to satisfy people who don't already know Grover's algorithm, since the module is more about its application, rather than an explanation about how Grover's algorithm is constructed? We can either add that in this tutorial or update the Grover's module or in the conceptual module. I'm also not sure, having learned the algorithm using circuits, if programmers find it easy to understand the algorithm without the aim of the circuit.

geduardo: The MS Learn module doesn't explain the implementation (that's the task of this tutorial), but it does explain why Grover's algorithm work (for example, with the rotation's diagrams).

About the circuits, I don't think we need them. You just need to have clear the steps of the algorithm. It could be useful tho a mathematical description of how the state of the register evolves during the algorithm.

KittyYeungQ: "It could be useful tho a mathematical description of how the state of the register evolves during the algorithm." Can you add this?

geduardo: Mm, I think that properly adding that here in this tutorial would make it too long and heavy for what's supposed to be a "quick implementation guide". But I also think that it could too mathematically dense for the MS Learn module.

What do you think about adding a conceptual article in the Concepts section (where we can use more "heavy maths") to explain properly Grover's algorithm? We can use it then as a reference for both, the MS Learn module and the implementation tutorial, for those users that want to dive deep into the math.

KittyYeungQ: Agreed. Adding it to the conceptual article makes sense.

geduardo: Ok! I'll write it up

PR #1137, opened 2020-11-27
Use correct simulator in ResourcesEstimator page

tcNickolas:


geduardo (approved):
Thanks! I also had headaches with that "s".

PR #1141, opened 2020-12-03
Update ms.topic metadata

Bradben:
Assigning the correct ms.topic values to each topic in order to better track SEO stats, page views, etc. Currently, most topics are assigned 'article', which is a valid generic value, but doesn't help with tracking.

The suggested default value for topics that don't fall into overview, tutorial, reference, quickstarts, etc., should be conceptual (as well as actual conceptual topics).


Bradben:
The namespaces for InvalidatedQubitsUseCheckerException and DistinctInputsCheckerException have been updated for the .NET API docs (csharp-quantum #43).  This PR should be ready to merge.

KittyYeungQ (approved):

PR #1142, opened 2020-12-04
Fixing redirecting issues

geduardo:


Bradben (approved):
Thanks for catching this!

PR #1146, opened 2020-12-07
Add index.md to IQ# magic commands TOC

### LINKED ISSUE #1145: IQ# magic commands don't have a list page
For Q# libraries APIs we have pages that list all operations in the namespace, such as [this one](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.math). It would be really convenient to add a similar page for IQ# magic commands, so that the person who's looking to see what commands exist can see all of them in one place with summaries. (Currently I couldn't find a way to share a list of magic commands with somebody, other than by pointing to one command and suggesting to look at others in the manu.)

rmshaffer:
This file is auto-generated. https://github.com/microsoft/iqsharp/pull/380 fixes the generating code, but we can also add the link here manually to fix this in the meantime, until the next QDK release. Fixes #1145.

Here's how the TOC looks with this change:
image

### DIFF
--- a/api/iqsharp-magic/toc.yml
+++ b/api/iqsharp-magic/toc.yml
@@ -1,3 +1,5 @@
+- href: index.md
+  name: 'IQ# magic commands'
 - href: azure.connect.md
   name: '%azure.connect magic command'
 - href: azure.execute.md


tcNickolas (approved):
Thank you!

PR #1150, opened 2020-12-19
Typo at PE without eigenstates

SoniaLopezBravo:
I think using the same dummy index in the denominator might be confusing. I propose to change that.


KittyYeungQ (approved):

KittyYeungQ:
Not sure why there are so many warnings from this build. @geduardo any suggestions?

Bradben:
> Not sure why there are so many warnings from this build. @geduardo any suggestions?

Most of these warnings have been fixed with my metadata check-in.

PR #1151, opened 2021-01-05
Conceptual grovers

geduardo:
This PR is to write the conceptual article for Grover's. After the review, it will be retargeted to the branch that is updating Grover's tutorial.


articles/tutorials/new_search.md
```diff
@@ -0,0 +1,476 @@
...
+if __name__ == "__main__":
+    main()
+
+```
+
+The program generates the following histogram:
+
+![alt_text=Histogram with the results of running several time the Grover's algorithm](~/media/grovers-histogram.png)
+
+As you can see in the histogram, the algorithm outputs the solutions to the search problem (1, 3, 7 and 21) with much higher probability than the non-solutions. You can think of Grover's algorithm as a quantum random generator that is purposefully biased towards those indexes that are a solutions of the search problem.
```
geduardo, line 472:
```suggestion
As you can see in the histogram, the algorithm outputs the solutions to the search problem (1, 3, 7 and 21) with much higher probability than the non-solutions. You can think of Grover's algorithm as a quantum random generator that is purposefully biased towards those indexes that are a solutions of the search problem. 
```

